PR #7, opened 2020-06-04
Register the right cmd_handle

ahcorde:
there's only 1 buffer for now, we should register the right command_handle


gazebo_ros2_control/src/default_robot_hw_sim.cpp
```diff
@@ -259,19 +259,23 @@ bool DefaultRobotHWSim::initSim(
     joint_states_[j] = hardware_interface::JointStateHandle(
       joint_names_[j], &joint_position_[j], &joint_velocity_[j], &joint_effort_[j]);
     if (register_joint_state_handle(&joint_states_[j]) != hardware_interface::HW_RET_OK) {
-      RCLCPP_WARN_ONCE(LOGGER, "cant register jointstatehandle");
+      RCLCPP_WARN_ONCE(LOGGER, "cant register joint_states_");
```
chapulina, line 262: Just a small note that a user who's reading this at runtime probably doesn't know what `joint_states_` is. Maybe something like `"Failed to register joint state handle."` could be more appropriate here.

Issue #10, opened 2020-06-05
Consider using gazebo_ros::Node instead of rclcpp:Node


chapulina:
I just realized `rclcpp::Node` is being used:

https://github.com/ros-simulation/gazebo_ros2_control/blob/98834ef700c1c6ffd3012d2de0a81a09ab2d2235/gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp#L64-L65

But if you used [gazero_ros::Node](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/ros2/gazebo_ros/include/gazebo_ros/node.hpp), you'd get the advantages listed [here](http://gazebosim.org/tutorials?tut=ros2_overview&cat=connect_ros) and [here](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-integration#migration-notes-3), so you don't need to worry about executors or parsing common SDF params.

ahcorde:
PR https://github.com/ros-simulation/gazebo_ros2_control/pull/11

PR #11, opened 2020-06-09
Used gazebo_ros::Node

ahcorde:
Related to this issue https://github.com/ros-simulation/gazebo_ros2_control/issues/10

### DIFF
--- a/gazebo_ros2_control/include/gazebo_ros2_control/gazebo_ros2_control_plugin.hpp
+++ b/gazebo_ros2_control/include/gazebo_ros2_control/gazebo_ros2_control_plugin.hpp
@@ -48,8 +48,6 @@
 #include "gazebo/common/common.hh"
 #include "gazebo/physics/physics.hh"
 
-#include "gazebo_ros/node.hpp"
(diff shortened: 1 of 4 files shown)


gazebo_ros2_control/include/gazebo_ros2_control/gazebo_ros2_control_plugin.hpp
```diff
@@ -48,6 +48,8 @@
 #include "gazebo/common/common.hh"
 #include "gazebo/physics/physics.hh"
 
+#include "gazebo_ros/node.hpp"
```
chapulina, line 51: This could be just included in the `cpp` file. It's not used by the header, so it shouldn't need to be exposed.

ahcorde: moved https://github.com/ros-simulation/gazebo_ros2_control/pull/11/commits/4912b176b41705bc70a6fd8a2ccfaf987d87de02

PR #19, opened 2020-06-24
Added CI

ahcorde:
Added CI

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,7 +27,8 @@ jobs:
           ros-foxy-control-msgs \
           ros-foxy-camera-info-manager \
           ros-foxy-cv-bridge \
-          ros-foxy-image-transport
+          ros-foxy-image-transport \
+          ros-foxy-ament-cmake-test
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,56 @@
...
+      id: build
+      run: |
+        cd /home/ros2_ws/
+        . /opt/ros/foxy/local_setup.sh
+        colcon build
+    - name: Run tests
+      id: test
+      run: |
+        cd /home/ros2_ws/
+        colcon test --event-handlers console_direct+ --packages-select gazebo_ros2_control
```
chapulina, line 56: One thing I learned recently is that `colcon test` doesn't return an error code if there are test failures. For that we need to run `colcon test-result` after running `colcon test` :scream: 

You can see on Travis that there are failures right now, even though the check is green :heavy_check_mark: 

```
The following tests FAILED:
	  1 - copyright (Failed)
	  2 - cppcheck (Failed)
	  3 - cpplint (Failed)
	  4 - lint_cmake (Failed)
	  5 - uncrustify (Failed)
	  6 - xmllint (Failed)
```

ahcorde: @chapulina  Fixed

ahcorde: Thank you for the tip!

PR #28, opened 2020-07-09
Update with ros2_control: return types (hardware_interface and controller_interface) 

ahcorde:
Return values have change in `hardware_interface` and `controller_interface` 

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -49,6 +49,7 @@ jobs:
       run: |
         cd /home/ros2_ws/
         . /opt/ros/foxy/local_setup.sh
+        export CMAKE_PREFIX_PATH=$AMENT_PREFIX_PATH:$CMAKE_PREFIX_PATH
         colcon build
     - name: Run tests
       id: test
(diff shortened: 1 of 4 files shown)


ahcorde:
@chapulina friendly ping

.github/workflows/ci.yaml
```diff
@@ -49,6 +49,7 @@ jobs:
       run: |
         cd /home/ros2_ws/
         . /opt/ros/foxy/local_setup.sh
+        export CMAKE_PREFIX_PATH=$AMENT_PREFIX_PATH:$CMAKE_PREFIX_PATH
```
bmagyar, line 52: I had to do the same after struggling a lot with my dev environment and don't get why. Could you please enlighten me?

ahcorde: This was addressed in [colcon-ros long time ago](https://answers.ros.org/question/296462/compilation-error-building-against-binary-bouncy-could-not-find-fastrtps/)  I think it's a missing dependency in one of the packages.

bmagyar: I've been chasing this for a while now, it's why I didn't want to release control_toolbox as I couldn't locally build it without applying hacks... I think it comes from control_msgs but it beats me what's missing...

ahcorde: these are `<exec_depend>` https://github.com/ros-controls/control_msgs/blob/foxy-devel/control_msgs/package.xml#L19-L22

You don't need these lines in the CMakeLists.txt https://github.com/ros-controls/control_msgs/blob/foxy-devel/control_msgs/CMakeLists.txt#L5-L9

I will have a look to the rest of the packages, where are you facing this issue?

PR #30, opened 2020-07-21
Upgrade packages in CI

ahcorde:
Upgrade packages in the CI but this still needs a release from rcpputils as requested [here](https://github.com/ros2/rcpputils/issues/89)


chapulina:
The CI error above is also happening upstream, see https://github.com/ros-controls/ros2_controllers/runs/901806756?check_suite_focus=true

I noticed that his repository doesn't need `joint_trajectory_controller` yet though, so with the change on fd13a8e I expect this PR to pass CI.

Issue #3, opened 2020-05-28
Reproduced current status of gazebo_ros2_control 


ahcorde:
I have created two PRs where I have added:

 - gazebo_ros2_control - preliminary version https://github.com/ros-simulation/gazebo_ros2_control/pull/1
 - gazebo_ros2_control_demos - preliminary demo package with a cart in a 30 meter rail. https://github.com/ros-simulation/gazebo_ros2_control/pull/2

The ROS 2 distro used for this tutorial is Foxy

## Run the demos:

```bash
mkdir -p ~/ros2_control_ws/src 
cd ~/ros2_control_ws/src
git clone https://github.com/ros-simulation/gazebo_ros2_control/ -b ahcorde/test
git clone https://github.com/ros/xacro -b dashing-devel
git clone https://github.com/ros-simulation/gazebo_ros_pkgs -b ros2
git clone https://github.com/ros-controls/realtime_tools -b ros2_devel
git clone https://github.com/ros-controls/ros2_control
git clone https://github.com/ros-controls/ros2_controllers
git clone https://github.com/ddengster/ros2_control/ -b coffeebot_deps ros2_control_ddengster
touch ros2_control_ddengster/COLCON_IGNORE
cp -r ros2_control_ddengster/joint_limits_interface ros2_control_ddengster/transmission_interface  ros2_control
cd ~/ros2_control_ws
colcon build --merge-install
```
### Running the examples

I created two examples: `hardware_interface/PositionJointInterface` and `hardware_interface/VelocityJointInterface`. I only made it work the `PositionJointInterface` example

Launch the Gazebo classic world with the gazebo_ros2_plugin
```
ros2 launch gazebo_ros2_control_demos cart_example_position.launch.py
```

Call the action `follow_joint_trajectories`:
```
ros2 run gazebo_ros2_control_demos example_position
```

Check `/joint_states` or `/state`

```
ros2 topic echo /joint_states
ros2 topic echo /state
```

![positionController](https://user-images.githubusercontent.com/1933907/83120541-7c2aa500-a0d1-11ea-9729-abee461e65b1.gif)

## Dependencies 
 - xacro
 - gazebo_ros_pkgs
 - ros2_control
 - ros2_controllers
 - realtime_tools

@chapulina @Karsten1987 @ddengster

oktkhr0916:
Thank you for your great project!

I have one question.
I tried the demo of `example_position` in my environment which is "ros2 eloquent".
[...]

PR #34, opened 2020-10-14
Updated with recent ros2_control changes

ahcorde:
Updated with recent ros2_control changes

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -29,10 +29,8 @@ jobs:
         git clone https://github.com/ros-controls/ros2_control
         git clone https://github.com/ros-controls/ros2_controllers
         git clone https://github.com/ddengster/ros2_control/ -b coffeebot_deps ros2_control_ddengster
-        git clone https://github.com/ros-controls/control_toolbox/ -b ros2-master
         touch ros2_control_ddengster/COLCON_IGNORE
-        cp -r ros2_control_ddengster/joint_limits_interface ros2_control_ddengster/transmission_interface  ros2_control
-        sed -i '/rclcpp/d' ros2_control/joint_limits_interface/CMakeLists.txt
+        cp -r ros2_control_ddengster/transmission_interface  ros2_control
         rosdep update
         rosdep install --from-paths ./ -i -y --rosdistro foxy \
           --ignore-src
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,9 +28,6 @@ jobs:
         cd /home/ros2_ws/src/
         git clone https://github.com/ros-controls/ros2_control
         git clone https://github.com/ros-controls/ros2_controllers
-        # git clone https://github.com/ddengster/ros2_control/ -b coffeebot_deps ros2_control_ddengster
-        # touch ros2_control_ddengster/COLCON_IGNORE
-        # cp -r ros2_control_ddengster/transmission_interface  ros2_control
         rosdep update
         rosdep install --from-paths ./ -i -y --rosdistro foxy \
           --ignore-src
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,9 +28,9 @@ jobs:
         cd /home/ros2_ws/src/
         git clone https://github.com/ros-controls/ros2_control
         git clone https://github.com/ros-controls/ros2_controllers
-        git clone https://github.com/ddengster/ros2_control/ -b coffeebot_deps ros2_control_ddengster
-        touch ros2_control_ddengster/COLCON_IGNORE
-        cp -r ros2_control_ddengster/transmission_interface  ros2_control
+        # git clone https://github.com/ddengster/ros2_control/ -b coffeebot_deps ros2_control_ddengster
+        # touch ros2_control_ddengster/COLCON_IGNORE
+        # cp -r ros2_control_ddengster/transmission_interface  ros2_control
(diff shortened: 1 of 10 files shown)


guru-florida:
Thanks @ahcorde!!!  It works!! I just couldnt seem to figure out how to get the startup node configuration right. I even tried to wrap my head around how the switch_controllers() call would work here but just couldnt see it. I still dont quite see it but I will dig into it this weekend.

Do you mind holding off on the PR for a bit? I had some fairly big updates as well, mostly based on ros2_control fixes but also some things I found while studying the code.
* We dont need to keep the local std::vector<double>s for all the joint states and commands anymore. The new RobotHardware class has register_joint() that create JointHandles directly on the DynamicJointState message.
* I have some changes that enable the joint saturation and soft limits including the velocity based ones. I moved the limits config code from init back into registerJointLimits() method where I saw some previous ifdef'd out code and updated with the new limit handles.
* added  comments

There  is a lot of code in default_robot_hw_sim and the logic is fairly complex, 90% of the complexity is in the setup init(). I had to write out a separate text doc to wrap my head around it all. I think a large chunk of the complexity is in the logic between the various control mode setup. How do you feel about me refactoring out the 3 control modes to separate classes? Like PositionJointXX, VelocityJointXX and EffortJointXX where XX is either Interface, Controller or nothing. I can create a draft PR and you can help review.

FYI  I also have a [PR pending in ros2_controls](https://github.com/ros-controls/ros2_control/pull/194) that will remove the shared_ptr parts of the joint handles. It can still work with shared_ptr but not required anymore.

guru-florida:
As per my [comment in #35](https://github.com/ros-simulation/gazebo_ros2_control/pull/35#issuecomment-711486331) I take back my request to hold off and I think this PR should move forward.

ahcorde:
Hi @guru-florida,

thank for your contibution! I have included your suggestions in the code. It's included/restored the pid controller.

Can you have a look ? If you want to make more modification please build on top of this PR.

ahcorde:
Regarding the refactoring, I'm fine with your suggested changes, but I think we should wait until the `ros2_control` and `ros2_controllers` API is more stable.

guru-florida:
Nice! Thank you!  Updates look good but I found a bug in one of my changes....the readSim() called in initSim() needs to be moved outside the loop at the end...readSim() iterates all joints and when number of joints >1 the readSim is called before other joints have been initialized. 

I believe there is a "components" update coming soon and hopefully things will stabilize after that. I'll probably be helping out conversion to components on the joint limits and transmission when that happens. I have a pending PR on transmission parsing too. These are basically the branches you stated in the #3 with updates to sync with latest ros2_controls master, uncrustification, linting, etc.

FYI I am working on getting my humanoid URDF up and simulating with Gazebo now.

ahcorde:
thank you @guru-florida for finding the bug!

Looking forward to see your humanoid! Any gif/image that you can share at this moment ?

Briancbn:
It seems that /joint_states and /states are not updating properly for the postion demo.

ahcorde:
@Briancbn  I can reproduce your issue. I'm looking into it

ahcorde:
@Briancbn  This last commit should fix the issue. Let me know if this works for you too.

guru-florida:
I just had a look. By the looks of your code change, it looks like you are re-fetching the handles for each of the joint interfaces in the readSim. Did you find the joint handles were being invalidated somehow? ....Those joint handles are pointing directly to doubles in the DynamicJointState msg....Oh @%$!! The ol' growth of std::vector reallocating memory and thus invalidating iterators before it. 

For now, I am going to test setting up the handles as normal, but delay the fetching of those references until after the initSim loop completes. At that point the msg vectors dont grow.

FYI This bug was also popping up in the limit handles enforce_limits() which also binds to those handles.

This IMO is a design flaw in RobotHardware and registering joint-handles as it presents unexpected behavior to the API user due to an inner implementation detail of std:vector. Possibly this will be solved with the new components work going into ros2_controls but we can bring it up in the next working group meeting.

Briancbn:
> @Briancbn This last commit should fix the issue. Let me know if this works for you too.

Thanks so much it works! Probably the best way tho, hahaha.

> FYI This bug was also popping up in the limit handles enforce_limits() which also binds to those handles.

I also notice this as well, for now I am just commenting this line out.

guru-florida:
@ahcorde The transmission parser code has been merged into ros2_control master. I have the changes to update this branch. I will issue a PR from my branch to this one.

guru-florida:
@ahcorde PR #38 is ready for merging into this branch. Also fixed up CI script to no longer do the ddengster branch checkout and overwrite. We have a gazebo plugin that runs directly on ros2_control masters!!! ....at least until they break it....and they will. :)

chapulina (commented):
It builds fine, and the position demo works for me.

But the velocity demo doesn't work. What I did:

```
ros2 launch gazebo_ros2_control_demos cart_example_velocity.launch.py &
# wait for simulation to launch
ros2 run gazebo_ros2_control_demos example_velocity
```

Am I missing something?

guru-florida:
@ahcorde I added PR #40 to fix the velocity controller example. Tested with position, position_pid and velocity examples.

@chapulina Thank you for taking the time to review.

ahcorde:
@guru-florida thank you for the fix ;)

@chapulina do you mind to try again ?

chapulina (approved):
Nice, all the demos work for me now! I didn't work closely at the code, but functionality-wise LGTM!

PR #35, opened 2020-10-18
Import/ahcorde update for ros2_controls and refactoring

guru-florida:
Primary goals of this PR is to:

1. Get the gazebo_ros2_control pkg working with the latest ros2_control changes
2. Reduce the complexity of imitSim() method - it does *a lot*

This code is based largely on the work of @ahcorde. Initially with his work in Issue #3, then in the recent PR #34. Big thanks for ahcorde figuring out the issue with the controller parameter loading sequencing...I struggled with this. The initSim() function for each joint loads the transmission and actuator URDF spec and configures ros2_control joint controllers, registers handles, hard/soft limits, links to gazebo joint instance and more.  This is a lot to take in for a single function. I think I've managed to simplify and reorganize the code for easier code readability without changing anything fundamental about the underlying behavior. 

**These are the things I have refactored**
* Moved all the validation logic to the top of the loop, mostly URDF input validation, to get it and the error reporting out of the way of actual control logic.
* Created a helper function (register_joint_interface lambda) to register joint state and command handles with useful error handling if registration fails or registered handles come back null. The handle registration for velocity, effort and position and commands are then simple one liners.
* registerJointLimits() method expanded to do all the limits config....I think it used to do this from the looks of the commented out code so I basically restored it's old capability. 
* New hard and soft joint limit handles in ros2_controls derive from a single base "joint limit" with common enforce_limits() method so I could simplify the joint limits code here as well. Now we only have a single joint_limit_handles array and registerJointLimits() creates the appropriate limit type (position, effort or velocity limiting type and soft vs hard).
* Lots of added comments

**What is not done:**
@ahcorde's PR #34 re-enabled PID control. I'd like to bring in these changes. 

My suggestion would be to look at the default_robot_hw_sim.h/cpp file, and specifically the initSim() function as a whole before looking at individual commits since the code is moved around a lot. I've validated the position card example still works as expected. @ahcorde I'd like to get your input and review on the changes.


guru-florida:
A lot of conflicting files I didnt touch since this is actually based on ahcorde/test branch...but no worries, I'll follow up with a rebase that should fix things up for the most part.

guru-florida:
After investigating the conflicts and differences between branches I see that the "test" branch was pretty far outdated and PR #34 already improves the readability of the initSIm code. I think PR#34 should move forward. If you agree, there are still some updates I can reapply to this branch but will reapply them on top of the PR34 branch.
* Adding more validation and error feedback
* Joint limit simplification
* remove arrays of doubles and use the new RobotHardware register_joint function that directly binds the handles to the DynamicJointState message

guru-florida:
Closing this PR since these changes have been integrated into PR #34.

PR #38, opened 2020-10-27
Ahcorde/update/ros2 control

guru-florida:
In a nutshell:
* fix errors due to names of members in the TransmissionInfo, ActuatorInfo, etc., interfaces were renamed in ros2_controls recent PR to be consistent with ros2_controls naming convention.
* moved declaration of PID parameters to precede the call to the PidROS constructor. PidROS looks for these parameters during construction and was causing SEGVs sometimes. In my debugger it seemed like the construction of the contained ParameterService was causing "parameter changed" events to fire and this was the source of the SEGVs. Also, the declaration of parameters didnt have defaults so log was complaining about parameters not being of floating type.
* To fix the issue of invalid handles, I split the joint loop in initSim into two. First registering handles, only then are handles retrieved and limits and PID can continue initializing.
* removed the GazeboRosControlPrivate parseTransmissionsFromURDF method. It was a 1 line call to the transmission parser and regardless of parsing it returned true so didnt handle errors. plugin Load now calls parser directly and catches errors.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,9 +28,9 @@ jobs:
         cd /home/ros2_ws/src/
         git clone https://github.com/ros-controls/ros2_control
         git clone https://github.com/ros-controls/ros2_controllers
-        git clone https://github.com/ddengster/ros2_control/ -b coffeebot_deps ros2_control_ddengster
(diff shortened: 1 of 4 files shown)


ahcorde:
Thank you @guru-florida !

PR #40, opened 2020-11-03
Fixed velocity controller example in Ahcorde/update/ros2 control

guru-florida:
Hi @ahcorde . This fixes the velocity controller example. Issue was due to declare_parameter default values pushing the velocity controller into PID control mode. Not having the defaults puts errors in the log output, so I also added a check for whether P, I and D joint parameters are actually set and I dont bother calling initPid() if not...this prevents the log error output.

Also, opmode handle was being registered twice by my last PR, removed the second duplicate one.


ahcorde (approved):
LGTM

PR #44, opened 2021-01-11
Updated with ros2-control Foxy API

ahcorde:
Updated gazebo_ros2_control with the ros2-control Foxy API 

 - joint_limit_interface is unavailable. See this for more details https://github.com/ros-controls/ros2_control/issues/279

NOTE: 

`ros-foxy-transmission-interface` is still not avaiblable, you should include this dependency in your workspace


ahcorde:
@bmagyar and @Karsten1987 do you mind to have a look ?

MartinPeris:
The demo for the velocity controller is not working. The position controller demo works properly.

Steps to reproduce this: 

- On a terminal execute: `ros2 launch gazebo_ros2_control_demos cart_example_velocity.launch.py`
- On another terminal execute: `ros2 run gazebo_ros2_control_demos example_velocity`

You will see that the cart is not moving. Digging around a bit, I saw that the node `velocity_test_node` is publishing to the topic `/commands` instead of `/cart_pole_controller/commands`. But even after remapping the topic to the correct one and making sure that the messages are published on the correct topic (`ros2 topic echo /cart_pole_controller/commands` shows the expected messages), the cart still doesn't move. Am I missing something?

ahcorde:
@MartinPeris thank you for testing. This commit should fix the issue https://github.com/ros-simulation/gazebo_ros2_control/pull/44/commits/a69ff7c25d4adc843fd5239c3310f181e5032ba2

MartinPeris:
Thanks for the quick fix! It works :)

JaehyunShim:
@ahcorde Before merging this PR to the master branch, I think those lines below should use RCLCPP_INFO rather than  RCLCPP_ERROR.

https://github.com/ros-simulation/gazebo_ros2_control/blob/ahcorde%2Fupdate%2Ffoxy/gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp#L289

https://github.com/ros-simulation/gazebo_ros2_control/blob/ahcorde%2Fupdate%2Ffoxy/gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp#L453

https://github.com/ros-simulation/gazebo_ros2_control/blob/ahcorde%2Fupdate%2Ffoxy/gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp#L499

https://github.com/ros-simulation/gazebo_ros2_control/blob/ahcorde%2Fupdate%2Ffoxy/gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp#L505

https://github.com/ros-simulation/gazebo_ros2_control/blob/ahcorde%2Fupdate%2Ffoxy/gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp#L513

https://github.com/ros-simulation/gazebo_ros2_control/blob/ahcorde%2Fupdate%2Ffoxy/gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp#L525

README.md
```diff
@@ -166,7 +173,9 @@ cart_pole_controller:
 
 #### Setting PID gains
 
-To set the PID gains for a specific joint you need to define them inside `<plugin><ros></plugin></ros>`. Using the generic way of defining parameters with `gazebo_ros`. The name of the parameter correspond the name of the joint followed by a dot and the name of the parameter: `p`, `i`, `d`, `i_clamp_max`, `i_clamp_min` and `antiwindup`.
+To set the PID gains for a specific joint you need to define them inside `<plugin><ros></plugin></ros>`.
```
bmagyar, line 176: Opening and closing tags are not in the right order here

Docker/Dockerfile
```diff
@@ -9,24 +9,34 @@ RUN apt-get update && apt-get install -q -y --no-install-recommends \
...
+    wget \
     python3-colcon-ros \
     && apt-get clean
 
 RUN mkdir -p /home/ros2_ws/src \
     && cd /home/ros2_ws/src/ \
-    && git clone https://github.com/ros-simulation/gazebo_ros2_control \
+    # && wget https://raw.githubusercontent.com/ros-planning/moveit2/main/moveit2.repos \
+    # && vcs import < moveit2.repos \
+    && git clone https://github.com/ros-planning/moveit2 \
```
chapulina, line 20: I think we can remove all these moveit demos now that the demos will live in https://github.com/ros-simulation/gazebo_ros_demos/pull/41 right?

gazebo_ros2_control/include/gazebo_ros2_control/gazebo_system.hpp
```diff
@@ -0,0 +1,140 @@
...
+    rclcpp::Node::SharedPtr & model_nh,
+    gazebo::physics::ModelPtr parent_model,
+    const urdf::Model * const urdf_model,
+    std::vector<transmission_interface::TransmissionInfo> transmissions,
+    sdf::ElementPtr sdf) override;
+
+  /// \brief callback to get the e_stop signal from the ROS 2 topic
+  void eStopCB(const std::shared_ptr<std_msgs::msg::Bool> e_stop_active);
+
+private:
```
chapulina, line 70: This header is installed. How about using the PIMPL pattern to hide these private members, so that the class can be changed without breaking ABI?

gazebo_ros2_control/include/gazebo_ros2_control/gazebo_system_interface.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+#include "transmission_interface/transmission_info.hpp"
+
+// URDF
+#include "urdf/model.h"
+
+namespace gazebo_ros2_control
+{
+
+class GazeboSystemInterface
```
chapulina, line 41: Some documentation could help users understand what this class is for

gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
```diff
@@ -220,16 +222,6 @@ void GazeboRosControlPlugin::Load(gazebo::physics::ModelPtr parent, sdf::Element
         impl_->control_period_.seconds());
   }
 
-  // Initialize the emergency stop code.
```
chapulina, line 223: Isn't this supported anymore? I still see `e_stop_topic` on the demo URDF

gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
```diff
@@ -244,49 +236,73 @@ void GazeboRosControlPlugin::Load(gazebo::physics::ModelPtr parent, sdf::Element
...
+
     impl_->executor_ = std::make_shared<rclcpp::executors::MultiThreadedExecutor>();
 
+    auto spin = [this]()
+      {
+        while (rclcpp::ok()) {
+          impl_->executor_->spin_once();
+        }
+      };
+    impl_->thread_executor_spin_ = std::thread(spin);
```
chapulina, line 286: Are you sure this is needed? The static `gazebo_ros::Node` that was got on line 138 should be using a global static executor.

gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
```diff
@@ -244,49 +236,73 @@ void GazeboRosControlPlugin::Load(gazebo::physics::ModelPtr parent, sdf::Element
...
       new controller_manager::ControllerManager(
-        impl_->robot_hw_sim_, impl_->executor_,
+        std::move(resourceManager_),
+        impl_->executor_,
         "gazebo_controller_manager"));
 
     // TODO(anyone): Coded example here. should disable when spawn functionality of
     // controller manager is up
-    auto load_params_from_yaml_node = [](rclcpp_lifecycle::LifecycleNode::SharedPtr lc_node,
+    auto load_params_from_yaml_node = [](rclcpp::Node::SharedPtr lc_node,
```
chapulina, line 298: I think the variable name should be changed to match the new type (i.e. no `lc`)

chapulina (commented):
I built `transmission_interface` from source and was able to run the demos :+1: 

We should update CI not to build `ros2_control` from source and use debs instead.

README.md
destogl:
```suggestion
To set the PID gains for a specific joint you need to define them inside `<plugin><ros></ros></plugin>`.
```

bmagyar:
I've just released `transmission_interface`, https://github.com/ros/rosdistro/pull/28054

gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
ahcorde: The controller manager needs this executor.

gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
ahcorde: yes, it's supported

gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
```diff
@@ -197,8 +201,10 @@ void GazeboRosControlPlugin::Load(gazebo::physics::ModelPtr parent, sdf::Element
   rclcpp::Duration gazebo_period(impl_->parent_model_->GetWorld()->Physics()->GetMaxStepSize());
```
Briancbn, line 201: This might not affect much but I realized that [`rclcpp::Duration`](https://docs.ros2.org/foxy/api/rclcpp/classrclcpp_1_1Duration.html) constructor is using **nanoseconds** while [`GetMaxStepSize()`](http://osrf-distributions.s3.amazonaws.com/gazebo/api/11.0.0/classgazebo_1_1physics_1_1PhysicsEngine.html#a0113a9c364f1d7fb9b2072161bec82fa) returns **seconds** in `double` format.

```suggestion
  rclcpp::Duration gazebo_period(impl_->parent_model_->GetWorld()->Physics()->GetMaxStepSize() * 1e9);
```

gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
```diff
@@ -197,8 +201,10 @@ void GazeboRosControlPlugin::Load(gazebo::physics::ModelPtr parent, sdf::Element
...
 
   // Decide the plugin control period
-  if (impl_->sdf_->HasElement("control_period")) {
-    impl_->control_period_ = rclcpp::Duration(impl_->sdf_->Get<double>("control_period"));
+  if (sdf->HasElement("control_period")) {
+    impl_->control_period_ = rclcpp::Duration(
```
Briancbn, line 205: I am not sure if this part needs to change. If control period is in terms of **seconds**, then the following should be fine
```suggestion
    impl_->control_period_ = rclcpp::Duration(sdf->Get<double>("control_period") * 1e9);
```

Briancbn (commented):
Hi @ahcorde, I was trying out this new version. I was trying to play around with the `control_period` some bugs were found.

Briancbn:
There are some additional strange behavior on my side, compared to `master` branch (with ros2_control and controllers set to respective version). The joints are not able to maintain at a certain position. I am using joint_state_controller and joint_trajectory_controller. the hardware interface is `hardware_interface/PositionJointInterface`.

I have attached the video here ([old](https://drive.google.com/file/d/1mFxvBt0eIqszysqLcDaDAFVODeYIiOIe/view?usp=sharing) )([new](https://drive.google.com/file/d/1mFxvBt0eIqszysqLcDaDAFVODeYIiOIe/view?usp=sharing))

I have also checked that the command that goes in `SetPosition()` inside `write()` is the correct value, however the state is not able to maintain the position. It seems to me it's how the communication between ros2_control and gazebo triggering this error.

ahcorde:
@Briancbn can you share your config files? Otherwise I can't reproduce it. Probably something wrong there because the rrbot is able to hold the joint positions.

Briancbn:
It's the same as the universal_robot one
controller.yaml
```yaml
joint_state_controller:
  ros__parameters:
    type: joint_state_controller/JointStateController
    publish_rate: 100

ur3e_pos_joint_traj_controller:
  ros__parameters:
    type: joint_trajectory_controller/JointTrajectoryController
    joints:
      - ur3e_shoulder_pan_joint
      - ur3e_shoulder_lift_joint
      - ur3e_elbow_joint
      - ur3e_wrist_1_joint
      - ur3e_wrist_2_joint
      - ur3e_wrist_3_joint
    constraints:
      goal_time: 0.6
      stopped_velocity_tolerance: 0.05
      ur3e_shoulder_pan_joint: {trajectory: 0.1, goal: 0.1}
      ur3e_shoulder_lift_joint: {trajectory: 0.1, goal: 0.1}
      ur3e_elbow_joint: {trajectory: 0.1, goal: 0.1}
      ur3e_wrist_1_joint: {trajectory: 0.1, goal: 0.1}
      ur3e_wrist_2_joint: {trajectory: 0.1, goal: 0.1}
      ur3e_wrist_3_joint: {trajectory: 0.1, goal: 0.1}
    stop_trajectory_duration: 0.5
    state_publish_rate: 500
    action_monitor_rate: 10

ur5_pos_joint_traj_controller:
  ros__parameters:
  ...
```

The xacro file, 
```xml
<gazebo>
    <plugin name="gazebo_ros2_control" filename="libgazebo_ros2_control.so">
        <robot_sim_type>gazebo_ros2_control/GazeboSystem</robot_sim_type>
        <parameters>$(find <package-name>)/config/controller.yaml</parameters>
    </plugin>
</gazebo>
```
Please do share the rrbot config with me as well just for reference. Thanks.

ahcorde:
> Could you please share the rrbot config with me as well just for reference. Thanks a lot!:)

Here it's the PR https://github.com/ros-simulation/gazebo_ros_demos/pull/41

C-monC:
ROS can't find rrbot_moveit_demo_nodes if I run the rocker or docker commands.

`rocker --x11 --nvidia --name gazebo_ros2_control_demo gazebo_ros2_control:latest`

`Executing command: 
docker run -it   --rm    --name gazebo_ros2_control_demo   --gpus all  -e DISPLAY -e TERM   -e QT_X11_NO_MITSHM=1   -e XAUTHORITY=/tmp/.docker.xauth -v /tmp/.docker.xauth:/tmp/.docker.xauth   -v /tmp/.X11-unix:/tmp/.X11-unix   -v /etc/localtime:/etc/localtime:ro    29e3892534f3 
Package 'rrbot_moveit_demo_nodes' not found: "package 'rrbot_moveit_demo_nodes' not found, searching: ['/home/ros2_ws/install', '/opt/ros/foxy']"
`
Does rocker check for packages outside it's container as well?

ahcorde:
The rrbot examples are not included in the Dockerfile, You should add the moveit repos and the rrbot packages, I think this [Dockerfile](https://github.com/ros-simulation/gazebo_ros2_control/blob/90a0b6f3fa39c8996c8173a9a3cacdaf916cd354/Docker/Dockerfile) may be usefull for you

Briancbn:
> Probably something wrong there because the rrbot is able to hold the joint positions.

I can verify that it does work properly on rrbot. You can go ahead and merge this. I will need to do some more investigation and get back to you, I am still not clear on what was causing my problem.

gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
```diff
@@ -386,7 +409,7 @@ void GazeboRosControlPlugin::Load(gazebo::physics::ModelPtr parent, sdf::Element
     YAML::Node root_node = YAML::LoadFile(impl_->param_file_);
```
v-lopez, line 409: At the moment the param file is defined as optional but in reality is mandatory.
If it is not provided the plugin crashes here.

For our use case it's better to load the controllers afterward, so we can reuse the same infrastructure for loading them in gazebo, other simulators and on the real robot.

gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
```diff
@@ -244,49 +243,73 @@ void GazeboRosControlPlugin::Load(gazebo::physics::ModelPtr parent, sdf::Element
...
+
     // Create the controller manager
-    RCLCPP_ERROR(impl_->model_nh_->get_logger(), "Loading controller_manager");
+    RCLCPP_INFO(impl_->model_nh_->get_logger(), "Loading controller_manager");
     impl_->controller_manager_.reset(
       new controller_manager::ControllerManager(
-        impl_->robot_hw_sim_, impl_->executor_,
+        std::move(resourceManager_),
+        impl_->executor_,
         "gazebo_controller_manager"));
```
v-lopez, line 301: Is there a strong reason to use "gazebo_controller_manager" instead of "controller_manager"?

The ros2_control cli tools assume "controller_manager", and although the node name can be specified, it makes all the commands less user friendly.

Example:
`ros2 control list_controllers` 
becomes
`ros2 control list_controllers --controller-manager gazebo_controller_manager`

gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
```diff
@@ -244,49 +243,73 @@ void GazeboRosControlPlugin::Load(gazebo::physics::ModelPtr parent, sdf::Element
...
 
     // TODO(anyone): Coded example here. should disable when spawn functionality of
```
v-lopez, line 303:
```suggestion
    impl_->executor_->add_node(impl_->controller_manager_);
    // TODO(anyone): Coded example here. should disable when spawn functionality of
```

Without adding the node the the executor, I cannot get the controller manager services to respond.

gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
```diff
@@ -396,17 +419,19 @@ void GazeboRosControlPlugin::Load(gazebo::physics::ModelPtr parent, sdf::Element
...
               impl_->param_file_,
               controller_name);
-            if (controller_name != "joint_state_controller") {
-              controller->get_lifecycle_node()->configure();
+            if (controller_name == "joint_state_controller") {
+              controller->assign_interfaces({}, std::move(loaned_state_ifs));
             }
+            controller->configure();
             start_controllers.push_back(controller_name);
           }
```
v-lopez, line 430: All this bit is not needed here anymore in my opinion.

I believe these operations should be done only inside the controller manager through the standard interface (the load and switch services).

gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
```diff
@@ -161,44 +165,49 @@ void GazeboRosControlPlugin::Load(gazebo::physics::ModelPtr parent, sdf::Element
...
   }
 
   // Get the robot simulation interface type
-  if (impl_->sdf_->HasElement("robot_sim_type")) {
-    impl_->robot_hw_sim_type_str_ = impl_->sdf_->Get<std::string>("robot_sim_type");
+  if (sdf->HasElement("robot_sim_type")) {
+    impl_->robot_hw_sim_type_str_ = sdf->Get<std::string>("robot_sim_type");
   } else {
-    impl_->robot_hw_sim_type_str_ = "gazebo_ros2_control/DefaultRobotHWSim";
+    impl_->robot_hw_sim_type_str_ = "gazebo_ros2_control/GazeboSystem";
```
Karsten1987, line 186: I do understand the logic within the `else` branch, that we are looking for a `gazebo_ros2_control/GazeboSystem` interface. But what does the `robot_sim_type` indicate?

gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
```diff
@@ -244,49 +243,73 @@ void GazeboRosControlPlugin::Load(gazebo::physics::ModelPtr parent, sdf::Element
     return;
   }
 
-  // Load the RobotHWSim abstraction to interface the controllers with the gazebo model
+  std::unique_ptr<hardware_interface::ResourceManager> resourceManager_ =
+    std::make_unique<hardware_interface::ResourceManager>();
```
Karsten1987, line 247: I deliberately don't highlight changes in style, but you're using camelCase here, whereas other variables use snake_case. Just fyi.

gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
```diff
@@ -244,49 +243,73 @@ void GazeboRosControlPlugin::Load(gazebo::physics::ModelPtr parent, sdf::Element
...
     urdf::Model urdf_model;
     const urdf::Model * const urdf_model_ptr =
       urdf_model.initString(urdf_string) ? &urdf_model : NULL;
 
     rclcpp::Node::SharedPtr node_ros2 = std::dynamic_pointer_cast<rclcpp::Node>(impl_->model_nh_);
-    if (!impl_->robot_hw_sim_->initSim(
-        impl_->model_nh_->get_namespace(), node_ros2, impl_->parent_model_, urdf_model_ptr,
-        impl_->transmissions_))
+    if (!gazeboSystem->initSim(
+        node_ros2,
```
Karsten1987, line 265: why does it need a node?

gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
```diff
@@ -244,49 +243,73 @@ void GazeboRosControlPlugin::Load(gazebo::physics::ModelPtr parent, sdf::Element
...
     }
 
+    std::vector<hardware_interface::LoanedStateInterface> loaned_state_ifs;
+    std::vector<hardware_interface::StateInterface> state_ifs;
+    state_ifs = gazeboSystem->export_state_interfaces();
+    for (auto state : state_ifs) {
+      loaned_state_ifs.emplace_back(hardware_interface::LoanedStateInterface(state));
+    }
+
+    resourceManager_->import_component(std::move(gazeboSystem));
```
Karsten1987, line 283: why are you manually exporting the state interfaces here? `import_component` should be all you need to use.

gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
```diff
@@ -396,17 +419,19 @@ void GazeboRosControlPlugin::Load(gazebo::physics::ModelPtr parent, sdf::Element
...
             load_params_from_yaml(
-              controller->get_lifecycle_node(),
+              controller->get_node(),
               impl_->param_file_,
               controller_name);
-            if (controller_name != "joint_state_controller") {
-              controller->get_lifecycle_node()->configure();
+            if (controller_name == "joint_state_controller") {
+              controller->assign_interfaces({}, std::move(loaned_state_ifs));
             }
```
Karsten1987, line 427: this seems very error prone. Why does the gazebo plugin have to deal with controllers? All that logic should be hidden within the controller manager.

gazebo_ros2_control/src/gazebo_system.cpp
```diff
@@ -0,0 +1,485 @@
...
+
+    // register the state handles
+    this->dataPtr->joint_pos_cmd_[j] = std::make_shared<hardware_interface::CommandInterface>(
+      joint_name, hardware_interface::HW_IF_POSITION, &this->dataPtr->joint_position_cmd_[j]);
+    this->dataPtr->joint_vel_cmd_[j] = std::make_shared<hardware_interface::CommandInterface>(
+      joint_name, hardware_interface::HW_IF_VELOCITY, &this->dataPtr->joint_velocity_cmd_[j]);
+    this->dataPtr->joint_eff_cmd_[j] = std::make_shared<hardware_interface::CommandInterface>(
+      joint_name, hardware_interface::HW_IF_EFFORT, &this->dataPtr->joint_effort_cmd_[j]);
+
+    try {
```
Karsten1987, line 240: as the transmissions are not fully baked yet, I would advise to leave out that logic. You're declaring quite a few parameters here on that node, which are not useful at the moment. 
Without that logic, I believe you also don't need a node pointer in the first place.

gazebo_ros2_control/src/gazebo_system.cpp
```diff
@@ -0,0 +1,485 @@
...
+  }
+};
+
+namespace gazebo_ros2_control
+{
+
+bool GazeboSystem::initSim(
+  rclcpp::Node::SharedPtr & model_nh,
+  gazebo::physics::ModelPtr parent_model,
+  const urdf::Model * const urdf_model,
```
Karsten1987, line 103: where is this used, and what happens if it's `NULL`?

Karsten1987 (commented):
I think there's a couple of places where we should re-iterate over. Namely the transmissions and the use of parameters/node usages within the plugin. 
I also think we shouldn't hardcode the `joint_state_controller` within the plugin. The controller manager/resource manager should be capable of handling all this and gazebo should never have any knowledge about loaded controllers. If the CM/RM are not able to handle this, there's something wrong on their side.

gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
v-lopez: This is a bit more complex than I thought because yaml files cannot be easily loaded into a running Node in ROS2.

The only way I've found is to load as part of a `Node` action in a launch file. If the Node is running, there's no `ros2 param load` or equivalent launch action to load a yaml into the param server.  https://github.com/ros2/ros2cli/issues/589

If I have time I'll try to add the ros2 param load, but I can't guarantee.

gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
ahcorde: In the documentation:

>  - `<robot_sim_type>`: The pluginlib name of a custom robot ros2_control system plugin to be used, defaults to `gazebo_ros2_control/GazeboSystem`

This is the plugin that you can rewrite based on your needs, you need to specify this in the plugin's parameters

gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
ahcorde: - logging
 - create the emergency stop subscriber
 - get the PID values (only for PID position)

gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
ahcorde: happy to remove some code

gazebo_ros2_control/src/gazebo_system.cpp
ahcorde: This was here to get the joints limit, but I think it could be removed

gazebo_ros2_control/src/gazebo_system.cpp
ahcorde: I don't get this one. Do you want me to remove the position and velocity PID ?

gazebo_ros2_control/src/default_robot_hw_sim.cpp
```diff
@@ -1,624 +0,0 @@
...
-                  nh_->get_logger(), "joint %s is configured in VELOCITY_PID mode",
-                  transmissions[j].joints[0].name.c_str());
-                break;
-            }
-          }
-        }
-      } catch (const std::exception & e) {
-        RCLCPP_ERROR(rclcpp::get_logger("gazebo_ros2_control"), "%s", e.what());
-      }
-      simjoint->SetParam("fmax", 0, joint_effort_limits_[j]);
```
Briancbn, line 312: @ahcorde This is making the difference for me. I comment out this line, then the main branch code is having the same strange behavior as this PR.

Does gazebo not respect the `effort` set inside `<limit effort=""/>` from the URDF? We still need this for position control.

ahcorde: Right now there is no `joint_limit_interface` available in `ros2_control`. There is a PR openned https://github.com/ros-controls/ros2_control/pull/271 to include this again in the `hardware_interace`.

Briancbn: Thanks. So the problem will not be there after `joint_limit_interface` is implemented. I previously didn't understand what was causing the behaviour that I was facing. You can resolve this thread now.

ahcorde:
As discussed yesterday on the `ros2_control` WG, I have removed the PID controllers. 

Would you mind taking another look?  @karsten @bmagyar

gazebo_ros2_control/include/gazebo_ros2_control/gazebo_system_interface.hpp
```diff
@@ -0,0 +1,67 @@
...
+#include "transmission_interface/transmission_info.hpp"
+
+// URDF
+#include "urdf/model.h"
+
+namespace gazebo_ros2_control
+{
+
+// SystemInterface provides API-level access to read and command joint properties.
+
```
Karsten1987, line 42:
```suggestion
```

gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
```diff
@@ -161,44 +169,49 @@ void GazeboRosControlPlugin::Load(gazebo::physics::ModelPtr parent, sdf::Element
...
   }
 
   // Get the robot simulation interface type
-  if (impl_->sdf_->HasElement("robot_sim_type")) {
-    impl_->robot_hw_sim_type_str_ = impl_->sdf_->Get<std::string>("robot_sim_type");
+  if (sdf->HasElement("robot_sim_type")) {
+    impl_->robot_hw_sim_type_str_ = sdf->Get<std::string>("robot_sim_type");
   } else {
-    impl_->robot_hw_sim_type_str_ = "gazebo_ros2_control/DefaultRobotHWSim";
+    impl_->robot_hw_sim_type_str_ = "gazebo_ros2_control/GazeboSystem";
```
Karsten1987, line 190: nitpick to highlight that it defaults to `GazeboSystem`:
```
impl_->robot_hw_sim_type_str_ = "gazebo_ros2_control/GazeboSystem";
if (sdf->HasElement("robot_sim_type")) {
    impl_->robot_hw_sim_type_str_ = impl_->sdf_->Get<std::string>("robot_sim_type");
}
```

gazebo_ros2_control/src/gazebo_system.cpp
```diff
@@ -0,0 +1,389 @@
...
+#include <vector>
+
+#include "gazebo_ros2_control/gazebo_system.hpp"
+
+class gazebo_ros2_control::GazeboSystemPrivate
+{
+public:
+  GazeboSystemPrivate() = default;
+
+  ~GazeboSystemPrivate() = default;
```
Karsten1987, line 26:
```suggestion

  ~GazeboSystemPrivate() = default;
```

gazebo_ros2_control/src/gazebo_system.cpp
```diff
@@ -0,0 +1,389 @@
...
+  }
+};
+
+namespace gazebo_ros2_control
+{
+
+bool GazeboSystem::initSim(
+  rclcpp::Node::SharedPtr & model_nh,
+  gazebo::physics::ModelPtr parent_model,
+  std::vector<transmission_interface::TransmissionInfo> transmissions,
```
Karsten1987, line 103: I don't want to hold this PR back, but I still don't understand what we are doing with transmissions here. In my opinion, there shouldn't be any `<transmission>` tag in the URDF any longer with the new ros2_control architecture. And as I've said previously, I don't think we have a clear design on how to parse/declare transmissions I advocate for leaving them out for now.

Karsten1987: Thinking further about this, I think the `configure` method should be sufficient enough to declare all parameters for the joint such as input method, PID values, et al. to provide all three control inputs (position, velocity, effort).

gazebo_ros2_control/src/gazebo_system.cpp
```diff
@@ -0,0 +1,389 @@
...
+          " has no associated joints.");
+      continue;
+    } else if (transmissions[j].joints.size() > 1) {
+      RCLCPP_WARN_STREAM(
+        this->nh_->get_logger(), "Transmission " << transmissions[j].name <<
+          " has more than one joint. Currently the default robot hardware simulation " <<
+          " interface only supports one.");
+      continue;
+    }
+    std::string joint_name = this->dataPtr->joint_names_[j] = transmissions[j].joints[0].name;
```
Karsten1987, line 154: you get all this data from the `hardware_info` in the `configure` method. No need for transmissions here.

gazebo_ros2_control/src/gazebo_system.cpp
```diff
@@ -0,0 +1,389 @@
...
+    } else if (joint_interfaces.size() > 1) {
+      // only a warning, allow joint to continue
+      RCLCPP_WARN_STREAM(
+        this->nh_->get_logger(), "Joint " << transmissions[j].name <<
+          " of transmission " << transmissions[j].name <<
+          " specifies multiple hardware interfaces. " <<
+          "Currently the default robot hardware simulation interface only supports one." <<
+          "Using the first entry");
+    }
+    std::string hardware_interface = joint_interfaces.front();
```
Karsten1987, line 185: this should be a parameter in the ros2_control tag and be parsed in the `configure` step accordingly.
c.f. https://github.com/ros-controls/ros2_control/blob/master/hardware_interface/include/hardware_interface/hardware_info.hpp

gazebo_ros2_control/src/gazebo_system.cpp
```diff
@@ -0,0 +1,389 @@
...
+    } else if (transmissions[j].joints.size() > 1) {
+      RCLCPP_WARN_STREAM(
+        this->nh_->get_logger(), "Transmission " << transmissions[j].name <<
+          " has more than one joint. Currently the default robot hardware simulation " <<
+          " interface only supports one.");
+      continue;
+    }
+    std::string joint_name = this->dataPtr->joint_names_[j] = transmissions[j].joints[0].name;
+
+    std::vector<std::string> joint_interfaces = transmissions[j].joints[0].interfaces;
```
Karsten1987, line 156: same comment as above, that info comes from the `ros2_control` tag, not the transmissions.

gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
```diff
@@ -244,49 +247,62 @@ void GazeboRosControlPlugin::Load(gazebo::physics::ModelPtr parent, sdf::Element
     return;
   }
 
-  // Load the RobotHWSim abstraction to interface the controllers with the gazebo model
+  std::unique_ptr<hardware_interface::ResourceManager> resource_manager_ =
+    std::make_unique<hardware_interface::ResourceManager>();
+
   try {
+    // We cannot use (for now) the <ros2_control> tag in the URDF, we need to call it manually.
```
Karsten1987, line 254: IMO, you do have access to it via the component parser: https://github.com/ros-controls/ros2_control/blob/master/hardware_interface/include/hardware_interface/component_parser.hpp

Karsten1987 (commented):
I don't want to hold this PR back from my end, so if anybody has a clear opinion that we should merge as is, please do.

my two cents here: 
I don't think the way the `gazebo_system` is initialized is the right way. I don't like to force a `<transmission>` tag in the URDF, when with the new control system we specify them in the ros2_control tag.
See [the design doc](https://github.com/ros-controls/roadmap/blob/master/design_drafts/components_architecture_and_urdf_examples.md) example 6.
```xml
  <ros2_control name="RRBotModularJoint1" type="actuator">
    <hardware>
      <plugin>ros2_control_demo_hardware/VelocityActuatorHardware</plugin>
      <param name="example_param_write_for_sec">1.23</param>
      <param name="example_param_read_for_sec">3</param>
    </hardware>
    <joint name="joint1">
      <command_interface name="velocity">
        <param name="min">-1</param>
        <param name="max">1</param>
      </command_interface>
      <state_interface name="velocity"/>
    </joint>
    <transmission name="transmission1">
      <plugin>transmission_interface/SimpleTansmission</plugin>
      <param name="joint_to_actuator">${1024/PI}</param>
    </transmission>
  </ros2_control>
```
To gather the information about which joints are available and which control interface is being used on them via the transmission tag seems to be a legacy of ROS1. You have all this information easily at hand with the component parser from the hardware interface. So you wouldn't need a transmission tag any longer in your URDF.

That being said, the Gazebo plugin in my eyes really only does the following:
* instantiate the resource manager/controller manager
* parse the ros2_control tag
* initialize the "gazebo_system_interface"
* add that interface to the resource manger
* `read`/`update`/`write`

README.md
```diff
@@ -96,8 +99,9 @@ The `gazebo_ros2_control` `<plugin>` tag also has the following optional child e
  - `<control_period>`: The period of the controller update (in seconds), defaults to Gazebo's period
  - `<robot_param>`: The location of the `robot_description` (URDF) on the parameter server, defaults to `robot_description`
  - `<robot_param_node>`: Name of the node where the `robot_param` is located, defauls to `robot_state_publisher`
- - `<robot_sim_type>`: The pluginlib name of a custom robot sim interface to be used, defaults to `gazebo_ros2_control/DefaultRobotHWSim`
+ - `<robot_sim_type>`: The pluginlib name of a custom robot ros2_control system plugin to be used, defaults to `gazebo_ros2_control/GazeboSystem`
  - `<parameters>`: YAML file with the configuration of the controllers
+ - `<e_stop_topic>`: Topic to publish the emergency stop
```
bmagyar, line 104: Is this a gazebo requirement? Is this a ros2_control requirement? We were planning to introduce some sort of e-stop handling but nothing is really designed for it yet. Is this to fulfill that use-case or a different one? I'd like to avoid having to remove functionality from the plugin/declare obsolete when we didn't really need to add it yet.

bmagyar (changes requested):
At this point I don't think I could add much without reiterating @Karsten1987 's main sentiment in https://github.com/ros-simulation/gazebo_ros2_control/pull/44#pullrequestreview-579627797 .

Just raised one question regarding emergency stop.

ahcorde:
@Karsten1987 I got rid of the `transmission_interface` package.

ahcorde:
@bmagyar removed the e-stop too

JaehyunShim:
@ahcorde 
[line 492](https://github.com/ros-simulation/gazebo_ros2_control/blob/ahcorde%2Fupdate%2Ffoxy/gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp#L492)  RCLCPP_ERROR  -> RCLCPP_INFO

gazebo_ros2_control/CMakeLists.txt
```diff
@@ -3,60 +3,68 @@ project(gazebo_ros2_control)
...
 
-add_library(default_robot_hw_sim SHARED src/default_robot_hw_sim.cpp)
-ament_target_dependencies(default_robot_hw_sim
-  angles
-  control_toolbox
-  gazebo_dev
+add_library(gazebo_hardware_plugins SHARED
+  src/gazebo_system.cpp
+)
+ament_target_dependencies(gazebo_hardware_plugins
```
Karsten1987, line 47: This line leads to undefined references locally. I needed to add `rclcpp` as a dependency to it.

ahcorde: Fixed https://github.com/ros-simulation/gazebo_ros2_control/pull/44/commits/595ba174d9f01403404160f659be5bd43f5fc7ce

Karsten1987:
Basically all demos crash with the following output:
```
 ➭ ros2 run gazebo_ros2_control_demos example_velocity
[INFO] [1612206600.790995219] [velocity_test_node]: node created
libc++abi.dylib: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
```

The cartpole is moving though within Gazebo, so I don't think that's a major issue.

The position example has a weird console output and I don't really understand what the pole is doing (it's bouncing a bunch of time left and right and stops at the origin again).
```
feedback->desired.velocities :-1
feedback->desired.positions :0.00146672
feedback->desired.velocities :-1
feedback->desired.positions :0.00146672
feedback->desired.velocities :-1
feedback->desired.positions :0.00146672
feedback->desired.velocities :-1
feedback->desired.positions :0.00146672
feedback->desired.velocities :-1
feedback->desired.positions :0.00146672
feedback->desired.velocities :-1
feedback->desired.positions :0.00146672
feedback->desired.velocities :-1
feedback->desired.positions :0.00146672
feedback->desired.velocities :-1
feedback->desired.positions :0.00146672
feedback->desired.velocities :-1
common_result_response time: 1612207082.215468
SUCCEEDED result code
async_send_goal
libc++abi.dylib: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
```

Karsten1987 (approved):
This looks ok from my end. I haven't had a chance to test the moveit integration and the rrbot example. Would be great if somebody could give that a shot as well to confirm.
@bmagyar @v-lopez

ahcorde:
I updated the `gazebo_ros_demos` ros2 branch to include the `ros2_control` tag.

JaehyunShim:
@ahcorde  it seems the current gazebo_ros2_control does not support multiple ros2_control tags. Only the joints in the first ros2_control tag are detected.

ahcorde:
@JaehyunShim  can you try the last commit https://github.com/ros-simulation/gazebo_ros2_control/pull/44/commits/794da79c0b46a8adc4a7ff909429a69d518eea99?

v-lopez:
I'm going to try this today, but for me https://github.com/ros-simulation/gazebo_ros2_control/pull/44#pullrequestreview-575263658 is still an issue. 

Just wrapping the code that parses the yaml file with:
```
if (!impl_->param_file_.empty())
{
  // Code that loads params
}
```
 is enough.

Also, last time I tried, I had error parsing complex yaml files like [this one](https://github.com/pal-robotics/pmb2_robot/blob/foxy-devel/pmb2_controller_configuration/config/mobile_base_controller.yaml). The double arrays were loaded as string arrays.

v-lopez:
I've tested our TIAGo mobile base with diff drive and it worked fine after some fixes to the diff drive controller.

ezgif-2-1b72b1bb33ae


We'll start working on a full TIAGo robot to test JTC and eventually MoveIt!. We're not full time on this, so I expect it to take a few weeks.

chapulina (commented):
The position and velocity demos work for me, but the effort demo is failing to find the `cart_pole_controller`, and that causes `gzserver` to crash:

```
[gzserver-1] [ERROR] [1612310632.481354867] [controller_manager]: Loader for controller 'cart_pole_controller' not found
[gzserver-1] terminate called without an active exception
[gzserver-1] Aborted (core dumped)
[ERROR] [gzserver-1]: process has died [pid 259898, exit code 134, cmd 'gzserver 
```

JaehyunShim:
1. Is there anything that can be done for consistency in the controller.yaml file format for [real physical robots](https://github.com/ros-controls/ros2_control_demos/blob/master/ros2_control_demo_robot/controllers/rrbot_forward_controller_position.yaml) and [simulated](https://github.com/ros-simulation/gazebo_ros2_control/blob/master/gazebo_ros2_control_demos/config/cartpole_controller.yaml) robots? 
2. Controllers are spawned in [urdf files](https://github.com/ros-simulation/gazebo_ros2_control/blob/master/gazebo_ros2_control_demos/urdf/test_cart_position.xacro.urdf#L62) now but can you make gazebo_ros2_control an executable so you can run it in a launch file as you could in [ROS1](https://github.com/ros-industrial/universal_robot/blob/kinetic-devel/ur_gazebo/launch/ur10.launch#L26)?

JaehyunShim:
> @JaehyunShim can you try the last commit [794da79](https://github.com/ros-simulation/gazebo_ros2_control/commit/794da79c0b46a8adc4a7ff909429a69d518eea99)?

It worked. Thanks!

JaehyunShim:
> There are some additional strange behavior on my side, compared to `master` branch (with ros2_control and controllers set to respective version). The joints are not able to maintain at a **non-zero** certain position. I am using joint_state_controller and joint_trajectory_controller. the hardware interface is `hardware_interface/PositionJointInterface`.
> 
> I have attached the video here ([old](https://drive.google.com/file/d/1mFxvBt0eIqszysqLcDaDAFVODeYIiOIe/view?usp=sharing) )([new](https://drive.google.com/file/d/1o5tg145gLgHDUtybG6uEZ64xaAFltsoI/view?usp=sharing)) Both models and controller parameters are exactly the same.
> 
> I have also checked that the command that goes in `SetPosition()` inside `write()` is the correct value, however the state is not able to maintain the position. It seems to me it's how the communication between ros2_control and gazebo triggering this error.
> 
> Update: links was wrong updated

I am encountering the same problem actually. How did you fix it?

Briancbn:
> I am encountering the same problem actually. How did you fix it?

@JaehyunShim Here is my comments https://github.com/ros-simulation/gazebo_ros2_control/pull/44#discussion_r565177746. There are several ways to workaround it. Currently, I created an additional gazebo plugin to help calling `SetParam("fmax", 0, <joint-force-limit>)` inside the `Load()` function.

>     * joint_limit_interface is unavailable. See this for more details [ros-controls/ros2_control#279](https://github.com/ros-controls/ros2_control/issues/279)

@ahcorde Should there be a notice in the doc on the problem that the missing`joint_limit_interface` may cause if you would like to make use of the `PositionInterface`, mainly due to **fmax** is unset?

v-lopez:
@JaehyunShim 
>     1. Is there anything that can be done for consistency in the controller.yaml file format for [real physical robots](https://github.com/ros-controls/ros2_control_demos/blob/master/ros2_control_demo_robot/controllers/rrbot_forward_controller_position.yaml) and [simulated](https://github.com/ros-simulation/gazebo_ros2_control/blob/master/gazebo_ros2_control_demos/config/cartpole_controller.yaml) robots?
> 
>     2. Controllers are spawned in [urdf files](https://github.com/ros-simulation/gazebo_ros2_control/blob/master/gazebo_ros2_control_demos/urdf/test_cart_position.xacro.urdf#L62) now but can you make gazebo_ros2_control an executable so you can run it in a launch file as you could in [ROS1](https://github.com/ros-industrial/universal_robot/blob/kinetic-devel/ur_gazebo/launch/ur10.launch#L26)?

Regarding this, we're not providing any ROS2 control params through the URDF, and instead use https://github.com/ros-controls/ros2_control/pull/310 to load them once gazebo and the controller manager in it are running. Here you can see the [controller params yaml file](https://github.com/pal-robotics/pmb2_robot/pull/5/files#diff-5153def005e3c975879d29aaae16a1f7a42bd31c6ca35b0a5dd566e67ba243bd)

I've got a launch file ready but not pushed, hopefully I can push it tonight.

ahcorde:
@JaehyunShim and  @v-lopez I unified the yaml file. Can you try it ?

ahcorde:
@chapulina, I was able to reproduce the issue in the docker container. The available classes are:

```
[gzserver-1] [ERROR] [1612347697.537043257] [controller_manager]: available classes:
[gzserver-1] [ERROR] [1612347697.537051929] [controller_manager]: controller_manager/test_controller
[gzserver-1] [ERROR] [1612347697.537056274] [controller_manager]: forward_command_controller/ForwardCommandController
[gzserver-1] [ERROR] [1612347697.537060021] [controller_manager]: joint_state_controller/JointStateController
[gzserver-1] [ERROR] [1612347697.537063475] [controller_manager]: joint_trajectory_controller/JointTrajectoryController
[gzserver-1] [ERROR] [1612347697.537067085] [controller_manager]: velocity_controllers/JointGroupVelocityController
[gzserver-1] [ERROR] [1612347697.537071095] [controller_manager]: Loader for controller 'effort_controllers' not found
```

I can see that the package `effort_controllers` [was released](https://github.com/ros/rosdistro/blob/master/foxy/distribution.yaml#L3009-L3032)  but it seems that it not still synchronized.

JaehyunShim:
@v-lopez I think @ahcorde made modifications on corresponding lines in the last commit for the first point. 

What I mean by the following
> 2. Controllers are spawned in [urdf files](https://github.com/ros-simulation/gazebo_ros2_control/blob/master/gazebo_ros2_control_demos/urdf/test_cart_position.xacro.urdf#L62) now but can you make gazebo_ros2_control an executable so you can run it in a launch file as you could in [ROS1](https://github.com/ros-industrial/universal_robot/blob/kinetic-devel/ur_gazebo/launch/ur10.launch#L26)?

is it is not really intuitive you set which controller you are going to use for your robot in a urdf file. You will have to recompile your package every time you switch to different controllers (or you can compile with the symlink option but still does not look that intuitive). Why does gazebo_ros2_control take this method rather than [link](https://github.com/ros-controls/ros2_control_demos/blob/master/ros2_control_demo_robot/launch/rrbot_system_position_only.launch.py#L41)?
,.

gazebo_ros2_control_demos/package.xml
```diff
@@ -33,7 +33,6 @@
   <exec_depend>robot_state_publisher</exec_depend>
```
chapulina, line 33: > I can see that the package effort_controllers was released but it seems that it not still synchronized.

Nice, I installed it from `ros2-testing` and now the demo works for me. Should we add it as an `exec_dependency` here?

chapulina (approved):
All the demos work well for me. The code also looks reasonable to me, although I'm not very familiar with the recent developments on `ros2_control`.

I suggest we get this in to fix the `master` branch and make a first release.  We can open tickets for all outstanding issues to be addressed in follow-up PRs. Otherwise we're going to feature-creep this pull request.

gazebo_ros2_control_demos/package.xml
ahcorde: Added an issue https://github.com/ros-simulation/gazebo_ros2_control/issues/49

JaehyunShim:
@Briancbn I still do not understand why I need to additionally add it to my code to make it work whereas the example provided in the gazebo_ros2_control_demos (which uses the same gazebo_ros2_control::GazeboSystem) already works without it.

Briancbn:
@JaehyunShim I only find out through experiment. I don't know what is causing this, and why the rrbot is working. Maybe @ahcorde or @chapulina could explain this?

ahcorde:
@JaehyunShim or @Briancbn can you open an issue ?

PR #45, opened 2021-01-12
Added moveit2 demo

ahcorde:
This PR builds on top of this other PR https://github.com/ros-simulation/gazebo_ros2_control/pull/44

I have created a simple example to move the joint of rrbot with moveit2 (which is already available in Foxy). Great new! Moveit2 and gazebo it's working! (more or less).

But I have some issues with the configuration files, for some reason the group is set to `panda_arm`. In the SRDF the group is called like this because I was not able to change it to something like `rrbot_arm`.

I'm able to move the robot but only to the defined position in the SRDF as you can see in the following gif.

rrbot_moveit

Any help with the configuration it's appreciate it @davetcoleman @henningkayser. Maybe these two packages should go in another repository.


swilcock0:
The [run_moveit_cpp](https://github.com/ros-planning/moveit2/blob/main/moveit_demo_nodes/run_moveit_cpp/src/run_moveit_cpp.cpp) demo is hardcoded for the panda arm hence why it won't let you rename the group. A modified version of the demo specific to the RRBot should work

ahcorde:
closing this PR in favour of this other https://github.com/ros-simulation/gazebo_ros_demos/pull/41

PR #55, opened 2021-02-08
Added License file

ahcorde:
Required to release the package https://github.com/ros/rosdistro/pull/28250

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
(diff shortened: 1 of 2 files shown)


gazebo_ros2_control/package.xml
```diff
@@ -7,6 +7,7 @@
   <maintainer email="ahcorde@osrfoundation.org">Alejandro Hernandez</maintainer>
 
   <license>BSD</license>
+  <license>Apache License 2.0</license>
```
jacobperron, line 10: It might be better to add two LICENSE files (one for each package), instead of a LICENSE file in the root directory, since this package is BSD and the other is Apache 2.

ahcorde: All files are Apache 2 except these two :

 - [gazebo_ros2_control_plugin.cpp](https://github.com/ros-simulation/gazebo_ros2_control/blob/79207cc7d7630a8aeced1b9f5bba985e0805e538/gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
 -  [gazebo_ros2_control_plugin.hpp](https://github.com/ros-simulation/gazebo_ros2_control/blob/79207cc7d7630a8aeced1b9f5bba985e0805e538/gazebo_ros2_control/include/gazebo_ros2_control_plugin.hpp)

But they have changed a lot since I started to port this plugin. Does it make sense to change the License to Apache 2 too (because of the complete redesing)?

jacobperron: If OSRF was the only copyright holder, I'd say it would be fine to change the license.

In this case, I can't recommend changing the license because it's not clear to me how influential the the original files were in developing this version.

I think the easiest thing to do would be to either,

1. License the entire package as BSD
2. Keep it dual licensed (list both BSD and Apache 2 in the package.xml and list both in the license file


IMO, option one is the simplest, considering an prospective user perspective. I'm open to other suggestions though.

ahcorde: Added BSD 3-Clause to the LICENSE file https://github.com/ros-simulation/gazebo_ros2_control/blob/7967bb7f957636315e48635d0774a6b2de96247a/LICENSE#L204-L234

dual License in the package.xml https://github.com/ros-simulation/gazebo_ros2_control/blob/7967bb7f957636315e48635d0774a6b2de96247a/gazebo_ros2_control/package.xml#L9-L10

LICENSE
```diff
@@ -0,0 +1,234 @@
...
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+
+----------
+
+BSD 3-Clause License
+
+Copyright (c) 2008, Willow Garage, Inc.
```
jacobperron, line 208: I might be mistaken, but I didn't think Willow was the original copyright holder.

jacobperron (approved):

PR #65, opened 2021-03-10
Add Imu and FT state interfaces

v-lopez:
Add state interfaces for gazebo sensors.

This allows sensors from being read directly from ROS control without going through ROS topics, which can better replicate real hardware where the sensor is read through the same bus as the actuators.

I took the liberty of establishing some standard interface names for IMU: "orientation.x", "linear_acceleration.x" and "angular_velocity.x". And obviously for y, z and w in the case of the orientation quaternion.

Maybe we could put this somewhere more standard , thoughts @bmagyar?

If this looks good, I'll do the same for the FT Sensor. 
We've been using this for years in ROS1 in https://github.com/pal-robotics/pal_hardware_gazebo/blob/melodic-devel/src/pal_hardware_gazebo.cpp but we think it will benefit more people here in ROS2.

### DIFF
--- a/gazebo_ros2_control/src/gazebo_system.cpp
+++ b/gazebo_ros2_control/src/gazebo_system.cpp
@@ -87,8 +87,17 @@ public:
   /// \brief An array per IMU with 4 orientation, 3 angular velocity and 3 linear acceleration
   std::vector<std::array<double, 10>> imu_sensor_data_;
 
-  /// \brief The current effort forces applied to the joints
+  /// \brief The ROS2 Control state interface for the current imu readings
   std::vector<std::shared_ptr<hardware_interface::StateInterface>> imu_state_;
+
+  /// \brief handles to the FT sensors from within Gazebo
+  std::vector<gazebo::sensors::ForceTorqueSensorPtr> sim_ft_sensors_;
+
+  /// \brief An array per FT sensor for 3D force and torquee
+  std::vector<std::array<double, 6>> ft_sensor_data_;
+
+  /// \brief The ROS2 Control state interface for the current FT sensor readings
+  std::vector<std::shared_ptr<hardware_interface::StateInterface>> ft_sensor_state_;
 };
 
 namespace gazebo_ros2_control
@@ -208,10 +217,32 @@ void GazeboSystem::registerSensors(
   // Collect gazebo sensor handles
   size_t n_sensors = hardware_info.sensors.size();
   std::vector<hardware_interface::ComponentInfo> imu_components_;
+  std::vector<hardware_interface::ComponentInfo> ft_sensor_components_;
+
+  // This is split in two steps: Count the number and type of sensor and associate the interfaces
+  // So we have resize only once the structures where the data will be stored, and we can safely
+  // use pointers to the structures
   for (unsigned int j = 0; j < n_sensors; j++) {
     hardware_interface::ComponentInfo component = hardware_info.sensors[j];
     std::string sensor_name = component.name;
-    std::vector<std::string> gz_sensor_names = parent_model->SensorScopedName(sensor_name);
+
+    // This can't be used, because it only looks for sensor in links, but force_torque_sensor
+    // must be in a joint, as in not found by SensorScopedName
+    // std::vector<std::string> gz_sensor_names = parent_model->SensorScopedName(sensor_name);
+
+    // Workaround to find sensors whose parent is a link or joint of parent_model
+    std::vector<std::string> gz_sensor_names;
+    for (const auto & s : gazebo::sensors::SensorManager::Instance()->GetSensors()) {
+      const std::string sensor_parent = s->ParentName();
+      if (s->Name() != sensor_name) {
+        continue;
+      }
+      if ((parent_model->GetJoint(sensor_parent) != nullptr) ||
+        parent_model->GetLink(sensor_parent) != nullptr)
+      {
+        gz_sensor_names.push_back(s->ScopedName());
+      }
+    }
     if (gz_sensor_names.empty()) {
       RCLCPP_WARN_STREAM(
         this->nh_->get_logger(), "Skipping sensor in the URDF named '" << sensor_name <<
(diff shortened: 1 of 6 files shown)


v-lopez:
On hold to review the naming of the interfaces as discussed on March 10th WG meeting. Sorry about the noise. But feel free to check the rest of the implementation as well.

Adapting something similar as the example in https://github.com/ros-controls/roadmap/blob/master/design_drafts/hardware_access.md#controller-interface

v-lopez:
We'll keep the current naming, feel free to give it a look, if it's fine I'll add the FT sensor in the same way.

v-lopez:
I just added ForceTorque sensor too.

I don't know if I should file an Issue for this: https://github.com/ros-simulation/gazebo_ros2_control/pull/65/files#diff-e14963d1c2ba3a31138bda99525946c355191fe361859b5fb7247fddad6c3a2aR229

A Force Torque sensor parent must be a Joint, but `Model::SensorScopedName` only looks for sensors in its links. So never finds them.

bmagyar:
@ahcorde @chapulina could we please get a review on this guys?

gazebo_ros2_control/src/gazebo_system.cpp
```diff
@@ -183,9 +204,145 @@ bool GazeboSystem::initSim(
...
+        RCLCPP_ERROR_STREAM(
+          this->nh_->get_logger(),
+          "Error retrieving casting sensor '" << sensor_name << " to ForceTorqueSensor");
+        continue;
+      }
+      ft_sensor_components_.push_back(component);
+      this->dataPtr->sim_ft_sensors_.push_back(ft_sensor);
+    }
+  }
+
```
ahcorde, line 291:
```suggestion
```

ahcorde (changes requested):
In general It looks good to me, but I would love to see any example world or documentation about how to use/interact with this.

gazebo_ros2_control/src/gazebo_system.cpp
jordan-palacios: Fixed.

jordan-palacios:
> In general It looks good to me, but I would love to see any example world or documentation about how to use/interact with this.

The idea is that the (WIP) IMU and FT sensor controllers (or broadcasters) in ros2_controllers can make use of the interfaces provided by the gazebo system plugin.

We will make use of this in our public robot simulations, though we need the support for it in the gazebo system plugin first.

jordan-palacios:
I forgot to mention that I will continue the work of this PR. :stuck_out_tongue:

bmagyar:
@jordan-palacios I think if you added an IMU to the cart and an extra statement to print the relevant values from flexible joint states, that would cover what Alex is asking for

jordan-palacios:
> @jordan-palacios I think if you added an IMU to the cart demo and an extra statement to the readme printing the relevant values from flexible joint states, that would cover what Alex is asking for

Good idea. Will do.

jordan-palacios:
I have added the IMU to the urdf of the `cart_example_position` demo. 

You can check the reported IMU data with:

`ros2 topic echo /dynamic_joint_states`

Unfortunately, something must be wrong with the my IMU configuration because the data ranges are all wrong.

For example, while idle, the cart reports a `linear_acceleration.x` of `-5.633648437341699e-59`. When the motion is triggered this value changes to, at best, to `2.9199594465887168e-31`. @ahcorde could you take a look please?

When we fix this I'll add the IMU to the remaining demos and document the usage in the README.

ahcorde:
@jordan-palacios 

In the position mode we are moving the box to the specific position with `setPosition()`, I'm not really sure that this will generate any kind of interaction with the IMU because there is no "real" movement.

Did you try with velocity ?

jordan-palacios:
> @jordan-palacios
> 
> In the position mode we are moving the box to the specific position with `setPosition()`, I'm not really sure that this will generate any kind of interaction with the IMU because there is no "real" movement.
> 
> Did you try with velocity ?

Just tried it (also pushed the changes for the velocity demo). Same issues. IMU moves, but in very low magnitudes.

v-lopez:
The IMU controller was merged on ros2_control and this could be merged here as well, you should be able to run the examples and see the imu changes.

bmagyar:
FYI Foxy and Rolling releases were also made with the imu broadcaster

gazebo_ros2_control_demos/urdf/test_cart_position.xacro.urdf
```diff
@@ -48,6 +48,34 @@
...
+    </inertial>
+    <!-- <visual>
+      <origin xyz="0 0 0"/>
+      <geometry>
+        <box size="0.1 0.1 0.1"/>
+      </geometry>
+      <material name="red">
+        <color rgba="0.8 0 0 1"/>
+      </material>
+    </visual> -->
```
ahcorde, line 64: Can we remove this ? or add a comment why this is usefull ?

gazebo_ros2_control_demos/urdf/test_cart_velocity.xacro.urdf
```diff
@@ -48,6 +48,35 @@
     <limit effort="1000.0" lower="-15" upper="15" velocity="30"/>
     <dynamics damping="0.0" friction="0.0"/>
   </joint>
+  <link name="imu">
+    <inertial>
+      <mass value="0.1"/>
+      <inertia ixx="1.0" ixy="0" ixz="0" iyy="1.0" iyz="0" izz="1.0"/>
+    </inertial>
+    <!-- <visual>
```
ahcorde, line 56: Same comment

gazebo_ros2_control_demos/config/cartpole_controller_velocity.yaml
```diff
@@ -8,6 +8,9 @@ controller_manager:
     joint_state_broadcaster:
```
ahcorde, line 8: do we need these changes in the position.yaml and position launch file ?

ahcorde (changes requested):

gazebo_ros2_control_demos/config/cartpole_controller_velocity.yaml
v-lopez: They are just an example, we only need them if we want the imu on those as well. But it doesn't make sense because afaik the cartpole moves instantaneously, so the IMU doesn't report anything useful.

gazebo_ros2_control_demos/urdf/test_cart_position.xacro.urdf
v-lopez: Sorry I'll remove the commented lines, they are not needed.

v-lopez: Also the whole imu for the position cart, I didn't understand

v-lopez:
Ping, is there anything missing?

ahcorde (approved):

roncapat:
Is there an example or a template to write an IMU ros2 control interface? The gazebo-ros2-control package is not a great start point if someone has to write a plugin for a real IMU hardware sensor from scratch.

PR #77, opened 2021-05-25
Ensure that sim_joints_ always has the same number of elements as the…

kbogert:
… joint_names

Otherwise a segfault occurs when some joints are not simulated by gazebo.

Possbily the cause of #67

### DIFF
--- a/gazebo_ros2_control/src/gazebo_system.cpp
+++ b/gazebo_ros2_control/src/gazebo_system.cpp
@@ -299,7 +299,6 @@ hardware_interface::return_type GazeboSystem::write()
 
   for (unsigned int j = 0; j < this->dataPtr->joint_names_.size(); j++) {
     if (this->dataPtr->sim_joints_[j]) {
-
       if (this->dataPtr->joint_control_methods_[j] & POSITION) {
         this->dataPtr->sim_joints_[j]->SetPosition(
           0, this->dataPtr->joint_position_cmd_[j],
--- a/gazebo_ros2_control/src/gazebo_system.cpp
+++ b/gazebo_ros2_control/src/gazebo_system.cpp
@@ -127,13 +127,13 @@ bool GazeboSystem::initSim(
     std::string joint_name = this->dataPtr->joint_names_[j] = hardware_info.joints[j].name;
 
     gazebo::physics::JointPtr simjoint = parent_model->GetJoint(joint_name);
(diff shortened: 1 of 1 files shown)


ahcorde (changes requested):
I'm not sure about adding a nullptr to the vector. This is happening because the joint_name doesn't exist, maybe it's because your configuration is wrong? or is there any specific use case for this ?

kbogert:
This is happening because there are joints in the urdf file that are not simulated by gazebo, there are lots of reasons why this could be the case (in our case, it's due to hacks needed to get parallel grippers to move correctly).

The alternative is to remove these joints from all of the vectors, reducing their size as the non-simulated joints are discovered.  This is do-able but much more complicated, and as far as I can tell having the nulls in the sim_joints vector isn't causing any issues.

gazebo_ros2_control/src/gazebo_system.cpp
```diff
@@ -283,6 +298,9 @@ hardware_interface::return_type GazeboSystem::write()
   rclcpp::Duration sim_period = sim_time_ros - this->dataPtr->last_update_sim_time_ros_;
 
   for (unsigned int j = 0; j < this->dataPtr->joint_names_.size(); j++) {
+    if (this->dataPtr->sim_joints_[j]) {
```
ahcorde, line 301: This indentation looks wrong

ahcorde (changes requested):

gazebo_ros2_control/src/gazebo_system.cpp
```diff
@@ -283,20 +298,23 @@ hardware_interface::return_type GazeboSystem::write()
...
-      this->dataPtr->sim_joints_[j]->SetVelocity(
-        0,
-        this->dataPtr->joint_velocity_cmd_[j]);
-    }
-    if (this->dataPtr->joint_control_methods_[j] & EFFORT) {
-      const double effort =
-        this->dataPtr->joint_effort_cmd_[j];
-      this->dataPtr->sim_joints_[j]->SetForce(0, effort);
+    if (this->dataPtr->sim_joints_[j]) {
+
```
ahcorde, line 302: remove this line, then it's good to go

ahcorde (changes requested):

ahcorde (changes requested):
@kbogert  uncrustify is failing

ahcorde (approved):

PR #79, opened 2021-05-28
Join with the controller manager's executor thread on exit

kbogert:
This prevents a SIGABRT in gzserver when stopping gazebo due to the executor thread being aborted, which triggers a core dump.  This patch instead ensures a clean exit, since rclcpp::ok() returns true while gazebo is shutting down.

Note: I'm not sure why the controller manager's node must be removed from the executor, but otherwise the join() will remain blocked and prevent gzserver from stopping even with a SIGTERM.


ahcorde (approved):
:heart:

PR #80, opened 2021-06-10
Forward sdf ros remappings to loaded controllers

jonatanolofsson:
Currently, the loaded controllers do not receive the remappings and namespace info from the sdf. This commit adds this information to the generated argument list so that they are forwarded to subsequently started nodes.

### DIFF
--- a/gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
+++ b/gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
@@ -200,6 +200,35 @@ void GazeboRosControlPlugin::Load(gazebo::physics::ModelPtr parent, sdf::Element
   // So we have to parse the plugin file manually and set it to the node's context.
   auto rcl_context = impl_->model_nh_->get_node_base_interface()->get_context()->get_rcl_context();
   std::vector<std::string> arguments = {"--ros-args", "--params-file", impl_->param_file_.c_str()};
+  if (sdf->HasElement("ros")) {
+    sdf = sdf->GetElement("ros");
+
+    // Set namespace if tag is present
+    if (sdf->HasElement("namespace")) {
+      std::string ns = sdf->GetElement("namespace")->Get<std::string>();
+      // prevent exception: namespace must be absolute, it must lead with a '/'
+      if (ns.empty() || ns[0] != '/') {
+        ns = '/' + ns;
+      }
+      std::string ns_arg = std::string("__ns:=") + ns;
+      arguments.push_back(RCL_REMAP_FLAG);
+      arguments.push_back(ns_arg);
+    }
+
+    // Get list of remapping rules from SDF
+    if (sdf->HasElement("remapping")) {
+      sdf::ElementPtr argument_sdf = sdf->GetElement("remapping");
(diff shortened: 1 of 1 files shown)


gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
```diff
@@ -200,6 +200,35 @@ void GazeboRosControlPlugin::Load(gazebo::physics::ModelPtr parent, sdf::Element
   // So we have to parse the plugin file manually and set it to the node's context.
   auto rcl_context = impl_->model_nh_->get_node_base_interface()->get_context()->get_rcl_context();
   std::vector<std::string> arguments = {"--ros-args", "--params-file", impl_->param_file_.c_str()};
+  if (sdf->HasElement("ros")) {
+    sdf = sdf->GetElement("ros");
+  }
```
ahcorde, line 205: This will allow the following

```xml
<plugin>
  <namespace></namespace>
  <remapping></remapping>
  ...
</plugin>
```

and 

```xml
<plugin>
  <ros>
    <namespace></namespace>
    <remapping></remapping>
    ...
 </ros>
</plugin>
```

I think we should only go with the second option

ahcorde (changes requested):

jonatanolofsson:
Hi, and thanks.

The behavior mirrors that of gazebo_ros node.cpp, which is the file that
reads the xml structure currently. While I agree a single standard would be
better, this follows the current behaviour more closely (although it might
be an unintended feature)


Best Regards,
Jonatan

jonatanolofsson:
But yes, I could support either way

jonatanolofsson:
@ahcorde Could you please advice if I should update the PR or not, based on the argument above?

ahcorde:
@jonatanolofsson yes, please I would prefer to update the PR

jonatanolofsson:
@ahcorde Done!

ahcorde (approved):

cco-seasony:
Hi @jonatanolofsson ,
I am trying to test this feature but for some reason things are not loading correctly.
When testing the namespace tag this is what I am encountering:

- I can load the plugin under the given namespace. I can see the services have the correct names.
- When trying to call `ros2 control load_start_controller velocity_controller` to start the velocity controller it does not work since that cmd is trying to use the service `/controller_manager/load_and_start_controller` which it does not exist since the namespace is prefixed to all the services. Hence, `/<namespace>/controller_manager/load_and_start_controller` is the one that this cmd`ros2 control load_start_controller velocity_controller` should be calling.

I can manually call the service `ros2 service call /<namespace>/controller_manager/load_and_start_controller controller_manager_msgs/srv/LoadStartController "{name: velocity_controller}"` and it works fine. This loads the controller and creates the command topic under the same namespace. 
I was wondering if this is the right way to do it or am I doing something wrong along the way?

Thanks in advance for the help.
Kind regards,
Carlos

jonatanolofsson:
@cco-seasony If you're using spawner.py, you can use the -c parameter to change which controller_manager it talks to. There might be a similar (same?) parameter for the ros2 control interface

PR #97, opened 2021-11-07
Pass ROS time instead of SYSTEM time to update function

bjsowa:
This fixes ros-controls/ros2_controllers#253
I haven't tested it with all the available controllers though.


bmagyar (approved):

bmagyar:
@ahcorde I think this should be merged

ahcorde (approved):

ahcorde:
should we need to backport this to foxy ?

bmagyar:
No, thank you. Passing time in is not part of the API in foxy.

PR #106, opened 2022-01-24
The Dockerfile in the main branch now uses galactic

aboarya:
The entry point needs to be updated to source /opt/ros/galactic instead of foxy

### DIFF
--- a/Docker/entrypoint.sh
+++ b/Docker/entrypoint.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
 
-. /opt/ros/foxy/setup.sh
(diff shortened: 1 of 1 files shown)


ahcorde (changes requested):
Can you retarget the PR to the `galactic` branch ?

`master` branch should be `rolling`.

aboarya:
> Can you retarget the PR to the `galactic` branch ?
> 
> `master` branch should be `rolling`.

of course! my mistake ...

ahcorde:
thank you @aboarya

PR #118, opened 2022-03-16
ros2_control is now having usings under its namespace.

destogl:
Check ros-controls/ros2_control#673 for more info. This breaks build of the plugin.﻿


destogl:
@ahcorde can you please review this PR because it is blocking rolling builds. (and please restart CI, I am not sure why this should be failing).

Changes are introduced in ros-controls/ros2_control#673

PR #122, opened 2022-03-31
support controller manager prefix and namespace for multiple controller managers

pabloinigoblasco:
This PR is created due to the need of having multiple controller managers to control multiple arms.

### DIFF
--- a/gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
+++ b/gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
@@ -196,6 +196,18 @@ void GazeboRosControlPlugin::Load(gazebo::physics::ModelPtr parent, sdf::Element
       impl_->model_nh_->get_logger(), "No parameter file provided. Configuration might be wrong");
   }
 
+  std::string controllerManagerNodeName{"controller_manager"};
+
+  std::string controllerManagerPrefixNodeName =
+    sdf->Get<std::string>("controller_manager_prefix_node_name");
+  if (!controllerManagerPrefixNodeName.empty()) {
+    controllerManagerNodeName = controllerManagerPrefixNodeName + "_" + controllerManagerNodeName;
+  }
+
+    std::string controllerManagerNodeNameNamespace{""};
+    controllerManagerNodeNameNamespace = sdf->Get<std::string>("controller_manager_namespace");
+
+
   // There's currently no direct way to set parameters to the plugin's node
   // So we have to parse the plugin file manually and set it to the node's context.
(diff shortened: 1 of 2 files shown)


ahcorde (changes requested):
Did you try to use `<plugin><ros><namespace></namespace></ros></plugin>` ?

```
<gazebo>
    <plugin filename="libgazebo_ros2_control.so" name="gazebo_ros2_control">
      <robot_param>...</robot_param>
      <robot_param_node>...</robot_param_node>
      <parameters>....</parameters>
      <ros>
         <namespace></namespace>
      </ros>
    </plugin>
</gazebo>

File: gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp [hunk 1]
```diff
@@ -196,6 +196,18 @@ void GazeboRosControlPlugin::Load(gazebo::physics::ModelPtr parent, sdf::Element
...
+  std::string controllerManagerNodeName{"controller_manager"};
+
+  std::string controllerManagerPrefixNodeName =
+    sdf->Get<std::string>("controller_manager_prefix_node_name");
+  if (!controllerManagerPrefixNodeName.empty()) {
+    controllerManagerNodeName = controllerManagerPrefixNodeName + "_" + controllerManagerNodeName;
+  }
+
+    std::string controllerManagerNodeNameNamespace{""};
+    controllerManagerNodeNameNamespace = sdf->Get<std::string>("controller_manager_namespace");
```
bmagyar, line 208:
```suggestion
  std::string controllerManagerNodeNameNamespace{""};
  controllerManagerNodeNameNamespace = sdf->Get<std::string>("controller_manager_namespace");
```
we should probably set up a clang-format file here too

gazebo_ros2_control/src/gazebo_system.cpp
```diff
@@ -27,6 +27,8 @@
 #include "hardware_interface/hardware_info.hpp"
 #include "hardware_interface/types/hardware_interface_type_values.hpp"
 
+using CallbackReturn = rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn;
```
bmagyar, line 30:
```suggestion
using CallbackReturn = rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn;
```


is this needed?

bmagyar (changes requested):

gazebo_ros2_control/src/gazebo_system.cpp
destogl: It should be needed, but probably under "namespace" scope and not global scope.

destogl:
Replaced by #147

PR #138, opened 2022-06-18
Adding simulation time parameter for the controller manager

eslamsalahelsheikh:
Adding the simulation parameter so that the controller manager uses the simulation time instead of the ROS time.  The '/odom' and corresponding tf will only be published if this parameter is set to true.

### DIFF
--- a/gazebo_ros2_control_demos/config/diff_drive_controller.yaml
+++ b/gazebo_ros2_control_demos/config/diff_drive_controller.yaml
@@ -1,6 +1,7 @@
 controller_manager:
   ros__parameters:
     update_rate: 100  # Hz
+    use_sim_time: true
 
     joint_state_broadcaster:
       type: joint_state_broadcaster/JointStateBroadcaster


destogl:
I don't understand this. This shouldn't affect the publication of anything. Are you sure that you are checking the correct setup?

If yes, please elaborate on the exact behavior.

eslamsalahelsheikh:
> I don't understand this. This shouldn't affect the publication of anything. Are you sure that you are checking the correct setup?
> 
> If yes, please elaborate on the exact behavior.

Hi Denis, 
the thing is that inside the diff_drive_controller there is a [check ](https://github.com/ros-controls/ros2_controllers/blob/d5507c3752ac8d5f16f9c96a9fe7be56cafd3535/diff_drive_controller/src/diff_drive_controller.cpp#L230)on the previous publishing time (you get it from [here](https://github.com/ros-controls/ros2_controllers/blob/d5507c3752ac8d5f16f9c96a9fe7be56cafd3535/diff_drive_controller/src/diff_drive_controller.cpp#L479)) and comparing it with the current time which is passed to the controller in this [line](https://github.com/ros-controls/ros2_controllers/blob/d5507c3752ac8d5f16f9c96a9fe7be56cafd3535/diff_drive_controller/src/diff_drive_controller.cpp#L142)  from the gazebo plugin [here](https://github.com/ros-controls/gazebo_ros2_control/blob/979e74269c2dd7dbb9ef231e2d1fa7626be22398/gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp#L377), and Gazebo is always using the [simulation time](https://github.com/ros-controls/gazebo_ros2_control/blob/979e74269c2dd7dbb9ef231e2d1fa7626be22398/gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp#L371) , so if the controller manager isn't using the same simulation time, this if condition will be always false and the [odom](https://github.com/ros-controls/ros2_controllers/blob/d5507c3752ac8d5f16f9c96a9fe7be56cafd3535/diff_drive_controller/src/diff_drive_controller.cpp#L246) and [tf](https://github.com/ros-controls/ros2_controllers/blob/d5507c3752ac8d5f16f9c96a9fe7be56cafd3535/diff_drive_controller/src/diff_drive_controller.cpp#L259) will not be published.

you can verify this easily if you run the [diff_drive example](https://github.com/ros-controls/gazebo_ros2_control/blob/master/gazebo_ros2_control_demos/launch/diff_drive.launch.py) with same setup you have [here](https://github.com/ros-controls/gazebo_ros2_control/blob/master/gazebo_ros2_control_demos/config/diff_drive_controller.yaml) and check if there is anything published on the "/odom" topic

destogl (approved):

PR #147, opened 2022-07-27
added namespace to controller manager

mcbed:
Minor fix to allow CM to be run in the defined namespace, ref #122


mcbed:
As `impl_->model_nh_->get_namespace()` returns the set namespace, do we still need to search for it in the sdf ?
https://github.com/ros-controls/gazebo_ros2_control/blob/a4bb81cb1b7707a8db5818329f1ba7528cf4e9c9/gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp#L205-L215

destogl (approved):

destogl:
@moutalibbadr @hugo-tardiou can you add this functionality to Ignition plugin, please? We should you exactly the same concept (or at least something very similar) there too.

ahcorde:
should we backport this @destogl ?

PR #136, opened 2022-06-15
implemented perform_command_mode_switch override in GazeboSystem

ksotebeer:
This PR should fix https://github.com/ros-controls/gazebo_ros2_control/issues/85 by implementing the `perform_command_mode_switch` method override to set joint control methods when controllers are activated/deactivated.

PR #150, opened 2022-08-15
[backport Foxy] #136 #147 #139 #134

ahcorde:
Backported
 - #136
 - #147
 - #139 
 - #134


ahcorde:
@bmagyar and @destogl let me know if we can backport something else to Foxy

PR #153, opened 2022-10-02
Add support for passive joints - joints with only state interfaces

destogl:
The PR if functional but the code is not tested against all use cases, especially those with mimic joints.
I don't think there are any issues, but this changes has to be extensivelly tested before merging.

### DIFF
--- a/gazebo_ros2_control/src/gazebo_system.cpp
+++ b/gazebo_ros2_control/src/gazebo_system.cpp
@@ -53,8 +53,11 @@ public:
   /// \brief last time the write method was called.
   rclcpp::Time last_update_sim_time_ros_;
 
-  /// \brief vector with the joint's names.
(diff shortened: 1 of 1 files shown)


christophfroehlich:
I tested this successfully with the basic examples in this repo, but what is the goal of this change @destogl? Passive joints with state interfaces only worked already before?

destogl:
> I tested this successfully with the basic examples in this repo, but what is the goal of this change @destogl? Passive joints with state interfaces only worked already before?

This PR separates commanding and state joints. The issue I was having was crashing of plugin when joint didn't have command interfaces, but only state interfaces. But you said this was working… Have to test this again then. I was using Gazebo Sim later for the project, so I am not sure anymore what the exact details behind it.

christophfroehlich:
See #172 : This is the way I used it since foxy up to now.. Never had problems with that.

Issue #156, opened 2022-10-31
/odom topic and tf between /odom and /base_link are not published in diff_drive.launch.py 


TatsukiNishimura:
My environments
- Ubuntu 20.04
- ROS2 galactic 
- Gazebo 11

When I run ``` diff_drive.launch.py ```  in gazebo_ros2_control_demos package, I try to see topics ```/odom``` and ```/tf``` with ``` ros2 topic echo ```, but no data was found from ```/odom``` and no tf data between ```/odom``` and ```/base_link```.

I suspect the solution is to set ```use_sim_time``` parameter of ```/controller_manager``` node to true.

destogl:
> I suspect the solution is to set use_sim_time parameter of /controller_manager node to true.

Did you test this solution?

PR #157, opened 2022-10-31
Fix/publish odom diff drive example

TatsukiNishimura:


TatsukiNishimura:
#156

ahcorde:
This was already fixed on `main`. I backported the fix https://github.com/ros-controls/gazebo_ros2_control/pull/158

PR #161, opened 2022-11-17
Add pre-commit hooks configuration

ahcorde:
Signed-off-by: Alejandro Hernández Cordero <ahcorde@gmail.com>

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,6 +12,10 @@ jobs:
       image: osrf/ros:foxy-desktop
     steps:
     - uses: actions/checkout@v2
+    - uses: actions/setup-python@v3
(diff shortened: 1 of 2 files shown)


destogl:
I would actually add this into a separate action, as we have in other ros2_control repositories. We can also for now ignore some of the check we are doing there, but in the long term it would make sense that we converge.

Check this files for the reference:
- https://github.com/ros-controls/ros2_control/blob/master/.pre-commit-config.yaml
- https://github.com/ros-controls/ros2_control/blob/master/.github/workflows/ci-format.yml

bmagyar (approved):
much good, very approved

destogl:
We have missed our belowed `.clang-format`

PR #168, opened 2022-12-28
Improve error message if robot_description_ param is wrong

christophfroehlich:
If the plugin is loaded with wrong parameters, e.g.
```urdf
      <robot_param>wrong_robot_description</robot_param>
      <robot_param_node>robot_state_publisher</robot_param_node>
```
the plugin crashed without console output of the segmentation fault. Now we get a more meaningful output.


christophfroehlich:
CI/build_testing should be fixed with #169

ahcorde (approved):

PR #177, opened 2023-02-03
Various bug fixes

AndyZe:
Just a straightforward port of this PR from `gz_ros2_control`:

https://github.com/ros-controls/gz_ros2_control/pull/114

### DIFF
--- a/gazebo_ros2_control/src/gazebo_system.cpp
+++ b/gazebo_ros2_control/src/gazebo_system.cpp
@@ -100,6 +100,9 @@ public:
 
   /// \brief mapping of mimicked joints to index of joint they mimic
   std::vector<MimicJoint> mimic_joints_;
+
+  /// \brief Gain which converts position error to a velocity command
+  double position_proportional_gain_;
 };
 
 namespace gazebo_ros2_control
@@ -123,6 +126,17 @@ bool GazeboSystem::initSim(
   if (physics_type_.empty()) {
     RCLCPP_ERROR(this->nh_->get_logger(), "No physics engine configured in Gazebo.");
     return false;
+  }
+
+  constexpr double default_gain = 0.1;
+  if (!this->nh_->get_parameter_or(
+      "position_proportional_gain",
+      this->dataPtr->position_proportional_gain_, default_gain))
+  {
+    RCLCPP_WARN_STREAM(
+      this->nh_->get_logger(),
+      "The position_proportional_gain parameter was not defined, defaulting to: " <<
+        default_gain);
   }
 
   registerJoints(hardware_info, parent_model);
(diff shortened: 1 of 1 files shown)


gwalck:
@AndyZe I am sorry for digging this merged solution to https://github.com/ros-controls/gz_ros2_control/issues/113 out, but I think the sagging issue is still there in gazebo classic after this back-ported fix #177. 

In POSITION control method my robot sags under gravity, and even worse, its joints slowly dislocate ! I built all `ros2_control` packages with `rolling` and latest `master` branch of `gazebo_ros2_control` and used some `kuka_experimental` robots.

https://github.com/ros-controls/gazebo_ros2_control/assets/132376938/f9d398c7-3da4-4154-9b51-8b801f3a0215

I was definitively sending a static command (with RQT JTC) to a JTC with position interface but the robot sags down. So its not the "robot not holding position" issue that is another recent issue of JTC it seems.

It reminded me of a well known problem from ROS and first use of gazebo 10 years ago to "enforce" a position that by-passes physics (= setting a pose or joint angle and not its effort).
We learnt the hard-way that one has to **not only** set **position** of a joint, but to also set its **velocity** to **zero** otherwise the non-zero velocity is still acting slowly overtime making the joint/object drift. 

You might think, this is a bug, because at the next cycle the position is set back to the correct command, and I would agree, but it seems since gazebo 2 this was never fixed or is a "feature". I don't know.

In any case, the solution we used in the past and which also works here to sagging in position mode (without a PID controller converting it to a velocity) is to also set velocity to zero.
I successfully tested it and can provide a PR.

More generally, I am not sure how this back-ported fix #177 helped with regard to sagging robots. I understand it in gz sim because there seems to be a POSITION_PID implemented and some velocity could be overwritten, but not in gz_classic version of the plugin which has no such thing (yet). 

old `gazebo_ros_control` in ROS had POSITION  and POSITION_PID  and VELOCITY and VELOCITY_PID etc... by the way. I don't know if an effort was made to simplify (retire PID versions) in `gazebo_ros2_control`.

bmagyar:
That being said at least we could have merged this wiy a better name on this repo but oh well :D

destogl:
Yes, please add the PR. This might fix UR simulation too

gwalck:
> Yes, please add the PR. This might fix UR simulation too

It does indeed fix UR sim. PR created

christophfroehlich:
Am I missing something, or is the parameter `position_proportional_gain_` not used any more?

ahcorde:
yes, it was added as a parameter, but it's not used in the code

christophfroehlich:
Should we then remove it again, or will it be used? ;)

ahcorde:
@christophfroehlich  https://github.com/ros-controls/gazebo_ros2_control/pull/220

Issue #185, opened 2023-03-13
Proposal: GPIO tags support, mapped to gazebo::Transport topics


roncapat:
This is a proposal (not sure if it's the right place for that) about adding to `gazebo_ros2_control` the possibility to read/write Gazebo topics. For example, in an use-case of mine, I needed an elegant way for my controller to set some kind of value to the (simulated) hardware, so to a Gazebo plugin. I checked the Gazebo API and found out that apart from introspection (read-only way to communicate) the other common solution (found in the PX4 SITL packages for example) is to bridge plugins via gazebo messages (Google protobufs).

Any interest in a function like that? Since I'm working on it already, it might be possible for me to provide a clean implementation (with some extra work).

ahcorde:
Happy to review the PR, I would suggest to include a tutorial or example too.

roncapat:
A possible work rationale here is to:
- 1 - interpret every command or state interface named differently from "position", "velocity", "effort" as write/read (pub/sub) proxies to gazebo topics.
- 2 - extend behavior to provide compatibility not only with `double` datatype but also at least `int` and `bool` (gazebo "Any" messages support different basic data types).
- 3 - introduce support to GPIOs (leveraging facilities developed in point 1 and 2)

Other desiderata?

PR #187, opened 2023-03-22
Gazebo transport bridge (#185, part 1)

roncapat:
This allows state_interfaces and command_interfaces with value different from "effort", "velocity" or "position" to be interpreted as Gazebo Transport topics to which either publish or subscribe.

Gazebo::Transport::Node class has to be redefined out-of-tree because the PR that introduces a required method addition has been submitted this same day (https://github.com/gazebosim/gazebo-classic/pull/3309), and thus is needed until upstream merge.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -46,7 +46,7 @@ jobs:
     container:
       image: osrf/ros:rolling-desktop
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Setup colcon workspace
       id: configure
       run: |
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -46,7 +46,7 @@ jobs:
     container:
       image: osrf/ros:rolling-desktop
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v2
     - name: Setup colcon workspace
       id: configure
       run: |
--- a/gazebo_ros2_control/CMakeLists.txt
+++ b/gazebo_ros2_control/CMakeLists.txt
@@ -26,7 +26,7 @@ link_directories(
 
 # Libraries
 add_library(${PROJECT_NAME} SHARED
-  src/gazebo_ros2_control_plugin.cpp
+  src/gazebo_ros2_control_plugin.cpp src/node.cc
 )
 ament_target_dependencies(${PROJECT_NAME}
   angles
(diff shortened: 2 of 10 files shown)


roncapat:
@ahcorde may I ask you to trigger the workflows? Not sure if the last one was triggered by you or not, but since it failed I made some fixes :)

roncapat:
Ok, up to now CI fails uncrustify/cpplint due to foreign code (Gazebo's modified node.hh & node.cc).

This means that:
- either we re-format the files (diverging from Gazebo's version)
- or we wait for merge of the gazebo PR and delete here the files

Any suggestions or ideas?

ahcorde:
Hi @roncapat, 

Thank you for your contribution but I would like to understand this better, what's the motivation behind this ? You already can access to these topics throught ROS 2 in particular the ros2_control APIs. Why do you need to expose this in gz::transport ?

roncapat:
> Hi @roncapat, 
> 
> Thank you for your contribution but I would like to understand this better, what's the motivation behind this ? You already can access to these topics throught ROS 2 in particular the ros2_control APIs. Why do you need to expose this in gz::transport ?

Hi, this is to use standalone, non-ROS gazebo plugins and to have access /expose over ROS2 (via ros_control) to all gazebo topics. So, both read/write capability. 

A note, if you tell me it's something you are going to consider, I have some more commits to push here then ;) Fixed some nasty bug (and started using in our lab environment with no further issues since then).

roncapat:
Linting of Gazebo-included files can't be fully achieved since cpplint and uncrustify collide each other on some lines. This may of course be avoided at all by https://github.com/gazebosim/gazebo-classic/pull/3309

bmagyar:
I've tried pushing a release just now but there's a missing requirement for Gazebo that the package.xml picks up by default. I don't think that has anything to do with your conflict, I suspect this PR is based on an ignition PR?

roncapat:
May I ask where I can see a log of the problem? From mobile I'm just seeing test failures related to linting.

roncapat:
@bmagyar may I ask you how would you like to proceed here / what I can do to revise this patch set?

christophfroehlich (changes requested):
What I often have to ask for :innocent: : Could you please add this feature to the documentation and/or add a new demo to the demo? Maybe then the use case gets clearer for everyone.

.github/workflows/ci.yaml
```diff
@@ -46,7 +47,7 @@ jobs:
     container:
       image: osrf/ros:rolling-desktop
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v2
```
christophfroehlich, line 50: why do you have to decrease the version here?

christophfroehlich:
> May I ask then also to you an opinion on this?

It seems that the gazebo-PR has a chance to get merged -> let's wait for that to avoid duplicating node.hh. Once gazebo is released, do we have to support older versions of gazebo from this package then? don't know..

Honestly speaking, I have too little experience to evaluate the benefits from this PR and therefore asked for a demo example.

gazebo_ros2_control/CMakeLists.txt
```diff
@@ -62,18 +64,8 @@ ament_export_include_directories(
   include
 )
 
-ament_export_dependencies(
-  ament_cmake
```
destogl, line 66: This is probably not intended change...

destogl (changes requested):
Please clean this up so we can do a proper review

PR #195, opened 2023-05-16
Fixed rolling compilation

ahcorde:
Related with this issue in the buildfarm https://build.ros2.org/job/Hdev__gazebo_ros2_control__ubuntu_jammy_amd64/32/console

we use the same solution in `gz_ros2_control`


gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
```diff
@@ -287,7 +287,11 @@ void GazeboRosControlPlugin::Load(gazebo::physics::ModelPtr parent, sdf::Element
   }
 
   for (unsigned int i = 0; i < control_hardware_info.size(); i++) {
+#ifndef ROLLING
+    std::string robot_hw_sim_type_str_ = control_hardware_info[i].hardware_class_type;
+#else
     std::string robot_hw_sim_type_str_ = control_hardware_info[i].hardware_plugin_name;
+    #endif
```
bmagyar, line 294: format!

PR #197, opened 2023-05-22
Fix deprecated copy warning

Noel215:
Hi,

Here's a simple change to remove a compilation warning that is complaining about a copy assignment for the type `SafeEnum`.

It's also happening in humble so if you could create a backport it would be awesome :)


bmagyar (approved):

bmagyar:
@mergifyio backport humble

PR #216, opened 2023-07-01
Add CI workflow to check docs

christophfroehlich:
This should add a CI job with the reusable workflow from https://github.com/ros-controls/control.ros.org/pull/125. It checks if there are no warnings from sphinx while building the docs.

### DIFF
--- a/.github/workflows/ci-check-docs.yml
+++ b/.github/workflows/ci-check-docs.yml
@@ -1,0 +1,12 @@
+name: Check Docs
+
+on:
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


saikishor (approved):

bmagyar (approved):

ahcorde:
do we need to backport this to humble ?

christophfroehlich:
I haven't tested that. I think if any PR to humble is created, it builds the humbe-PR as part of the master-version of control.ros.org, what should serve the purpose, too.

PR #223, opened 2023-08-14
Activate CI for humble

christophfroehlich:
CI is not activated for branches other than master:

[Gazebo ros2 control CI](https://github.com/ros-controls/gazebo_ros2_control/actions/workflows/ci.yaml)

I don't think this is on purpose?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -3,7 +3,7 @@ name: Gazebo ros2 control CI
 on:
   pull_request:
   push:
-    branches: 
+    branches:
       - master
       - humble
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -3,7 +3,9 @@ name: Gazebo ros2 control CI
 on:
   pull_request:
   push:
-    branches: [ master ]
+    branches: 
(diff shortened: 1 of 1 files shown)


ahcorde:
we have different branches for each one of the ROS version, this might fail because of the different `ros_control` APIs

christophfroehlich:
I'm aware of the branches, but the workflow file in the humble branch doesn't run on pushs on humble:
https://github.com/ros-controls/gazebo_ros2_control/blob/8ce87338b8fae820f0877ec68532b47d72fed005/.github/workflows/ci.yaml#L6

maybe the workflow file doesn't have to be updated on the master branch, but it works if they get updated on both as far as I know.

christophfroehlich:
I made this PR because humble has no status in the table of the readme
image

bmagyar (approved):

destogl (approved):

christophfroehlich:
apparently, it still doesn't work with the change on the master branch. Can we try a backport of this change to humble?

ahcorde:
https://github.com/Mergifyio backport humble

christophfroehlich:
now it's green, thanks ;)
image

Issue #235, opened 2023-09-14
Unexpected behavior when using passive joints in closed-loop kinematic chains


huemerj:
Probably related to #219, I experienced problems using passive joints when updating from 0.4.3 to 0.4.4 on humble.
More specifically, I simulate an actively controlled hydraulic cylinder acting on a passive joint in a closed-loop kinematic chain, which I need to get the state from.
As in #219, the passive joint is now fixed, even if it should be moving.

I narrowed down the problem to this part of the code, introduced in #177:

https://github.com/ros-controls/gazebo_ros2_control/blob/0dc19ab5561f7ef6b6d9addee55a4b01c56de126/gazebo_ros2_control/src/gazebo_system.cpp#L597-L600

destogl:
Thanks for reporting the issue. The issue is that we don't filter the passive joints. I propose to add a new map or vector with flags is joint is passive or not, and if it is passive, then we don't set the velocity here. A good place to fill this vector is after all interfaces are parsed on the line [296](https://github.com/ros-controls/gazebo_ros2_control/blob/0dc19ab5561f7ef6b6d9addee55a4b01c56de126/gazebo_ros2_control/src/gazebo_system.cpp#L296). A passive joits can be detected that there a no command interfaces, but only state interfaces.

Issue #219, opened 2023-08-11
Passive joints are broken


christophfroehlich:
As a consequence of #177 , passive joints don't work any more (try #172).

The joint is now fixed, even if no command interface is defined.

bmagyar:
That PR was fairly old I think when we merged it, it's not too surprising something wasn't fully aligned.

bmagyar:
I'll make myself a reminder note to never merge wildly unscoped PRs in the future ;)

christophfroehlich:
fixed with #237

PR #237, opened 2023-10-04
Fix stuck passive joints

huemerj:
As mentioned in #235 and #219, a problem with passive joints was introduced by the following code line: 
https://github.com/ros-controls/gazebo_ros2_control/blob/0dc19ab5561f7ef6b6d9addee55a4b01c56de126/gazebo_ros2_control/src/gazebo_system.cpp#L597-L600
As suggested by @destogl, I added a vector with flags which indicates if a joint is active (has command interfaces) or passive and subsequently sets the velocity to zero or not.

### DIFF
--- a/gazebo_ros2_control/src/gazebo_system.cpp
+++ b/gazebo_ros2_control/src/gazebo_system.cpp
@@ -59,6 +59,9 @@ public:
   /// \brief vector with the control method defined in the URDF for each joint.
   std::vector<GazeboSystemInterface::ControlMethod> joint_control_methods_;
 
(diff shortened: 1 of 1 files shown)


ahcorde (approved):
Do you have any plans to port this to `gz_ros_control`? If you don't have time, I can do it

ahcorde:
https://github.com/Mergifyio backport humble

huemerj:
> Do you have any plans to port this to `gz_ros_control`? If you don't have time, I can do it

I will have to set up / convert an example with passive joint for gz_sim, but I can do that.

Issue #250, opened 2023-12-31
Robot still saggs if interfaces are not claimed


christophfroehlich:
Since #177 the joints without claimed interfaces are set to have zero velocity. I'm still not exactly sure why this was introduced, because IMHO it does not work but breaks other configurations (passive joints were already fixed; but joints without active controller might be expected to fall down, e.g. with effort interface): 

- `ros2 launch ros2_control_demo_example_9 rrbot_gazebo_classic.launch.py gui:=true`
- Change the gravity to be horizontal in gazebo UI
- The robot moves

https://github.com/ros-controls/gazebo_ros2_control/assets/3367244/be3d1751-a2f8-4d3f-91fb-89f5d680c9db

If this is an important usecase, we should save the position at time of initialization or perform_command_mode_switch (if a controller was deactivated later) and set also the position to that value, not only the velocity to zero (similar to #213)
https://github.com/ros-controls/gazebo_ros2_control/blob/3c8f5bb4a4cc5eb01240afdbda9094f3ea3263ec/gazebo_ros2_control/src/gazebo_system.cpp#L604-L607

PR #251, opened 2023-12-31
Add `hold_joints` parameter

christophfroehlich:
@HPCLOL posted a use case of gazebo in https://github.com/ros-controls/ros2_controllers/issues/875:
If the simulation is started with an inactive controller, the robot should move with gravity as if the joints were passive.

As discussed with #250 this is currently not possible. This PR adds an parameter to enable this behavior (again). 

- I'm not sure if this should be default true or false.
- I haven't found a different solution to tell `GazeboSystem` this parameter parsed from the gazebo plugin without changing API. Therefore, I used a ROS parameter of the gazebo_ros2_control node to pass the information.

The result looks like this if no controller was spawned:

https://github.com/ros-controls/gazebo_ros2_control/assets/3367244/45e2f56a-7173-4cee-a30e-02713127c63b

### DIFF
--- a/gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
+++ b/gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
@@ -314,8 +314,20 @@ void GazeboRosControlPlugin::Load(gazebo::physics::ModelPtr parent, sdf::Element
       robot_hw_sim_type_str_.c_str());
     try {
       node_ros2->declare_parameter("hold_joints", rclcpp::ParameterValue(hold_joints));
-    } catch (const std::exception & e) {
-      RCLCPP_ERROR(impl_->model_nh_->get_logger(), "%s", e.what());
+    } catch (const rclcpp::exceptions::ParameterAlreadyDeclaredException & e) {
+      RCLCPP_ERROR(
+        impl_->model_nh_->get_logger(), "Parameter 'hold_joints' has already been declared, %s",
+        e.what());
+    } catch (const rclcpp::exceptions::InvalidParametersException & e) {
+      RCLCPP_ERROR(
+        impl_->model_nh_->get_logger(), "Parameter 'hold_joints' has invalid name, %s", e.what());
+    } catch (const rclcpp::exceptions::InvalidParameterValueException & e) {
+      RCLCPP_ERROR(
+        impl_->model_nh_->get_logger(), "Parameter 'hold_joints' value is invalid, %s", e.what());
+    } catch (const rclcpp::exceptions::InvalidParameterTypeException & e) {
+      RCLCPP_ERROR(
+        impl_->model_nh_->get_logger(), "Parameter 'hold_joints' value has wrong type, %s",
+        e.what());
     }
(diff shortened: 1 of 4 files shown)


saikishor (approved):
Thanks for the change! LGTM!

After discussion with @christophfroehlich 
We can add a new option for now, as we are limited to one control mode per run of the simulation. Ultimately, I think it will make sense to be able to switch the control modes dynamically once the simulation starts. We can try to aim this for Jazzy. If we can get this switching done, then we can remove this option, as we can achieve the same by switching the control mode to effort.

Ideally, the behavior would be at startup we will have the robot maintaining the position, and then when you switch from position or velocity active controllers to inactive mode, they maintain the old reference, and for the effort based controller, the reference will be set to zero.

What do you think about this new approach for future?

gazebo_ros2_control/src/gazebo_system.cpp
```diff
@@ -128,6 +131,18 @@ bool GazeboSystem::initSim(
     return false;
   }
 
+  try {
+    this->dataPtr->hold_joints_ = this->nh_->get_parameter("hold_joints").as_bool();
+  } catch (const std::exception & e) {
```
ahcorde, line 136: Include exception

gazebo_ros2_control/src/gazebo_ros2_control_plugin.cpp
```diff
@@ -306,6 +312,11 @@ void GazeboRosControlPlugin::Load(gazebo::physics::ModelPtr parent, sdf::Element
     RCLCPP_DEBUG(
       impl_->model_nh_->get_logger(), "Loaded hardware interface %s!",
       robot_hw_sim_type_str_.c_str());
+    try {
+      node_ros2->declare_parameter("hold_joints", rclcpp::ParameterValue(hold_joints));
+    } catch (const std::exception & e) {
```
ahcorde, line 317: Include exception

ahcorde (changes requested):

christophfroehlich:
@ahcorde not exactly sure what you requested, I added now explicit catch statements for the exceptions written in the docstring of the rclcpp functions. Or do you have another resource on how to write that properly?

ahcorde:
> @ahcorde not exactly sure what you requested, I added now explicit catch statements for the exceptions written in the docstring of the rclcpp functions. Or do you have another resource on how to write that properly?

My suggestion was about

```cpp
#include <exception>
```

But this change is also fine

ahcorde (approved):

christophfroehlich:
> > @ahcorde not exactly sure what you requested, I added now explicit catch statements for the exceptions written in the docstring of the rclcpp functions. Or do you have another resource on how to write that properly?
> 
> My suggestion was about
> 
> ```c++
> #include <exception>
> ```
> 
> But this change is also fine

Ah :grin: I thought too complicated. Thx!

christophfroehlich:
@ahcorde Could we backport this please? rolling won't be released any more.

christophfroehlich:
@mergifyio backport humble iron

PR #261, opened 2024-01-10
Rolling CI add sign from ROS2 packages

ahcorde:
I saw these [error messages](https://github.com/ros-controls/gazebo_ros2_control/actions/runs/7476086437/job/20345684887?pr=255) when building rolling
```bash
W: GPG error: http://packages.ros.org/ros2-testing/ubuntu jammy InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY F42ED6FBAB17C654
E: The repository 'http://packages.ros.org/ros2-testing/ubuntu jammy InRelease' is not signed.
```


christophfroehlich (approved):
I also noticed them, but haven't looked into it..
LGTM, thx!

PR #271, opened 2024-01-24
Fix crashing due to an invalid parameter in the initial value.

Wiktor-99:
Fix crashes caused by the wrong initial value in the state interfaces. This could be particularly problematic if someone passes 'pi' in xacro instead of '${pi}'. Xacro will evaluate '${pi}' to '3.14', while 'pi' will be evaluated as is ('pi' -> 'pi'), leading to chrash.

### DIFF
--- a/gazebo_ros2_control/src/gazebo_system.cpp
+++ b/gazebo_ros2_control/src/gazebo_system.cpp
@@ -233,14 +233,24 @@ void GazeboSystem::registerJoints(
 
     RCLCPP_INFO_STREAM(this->nh_->get_logger(), "\tState:");
 
-    auto get_initial_value = [this](const hardware_interface::InterfaceInfo & interface_info) {
+    auto get_initial_value =
+      [this, joint_name](const hardware_interface::InterfaceInfo & interface_info) {
+        double initial_value{0.0};
         if (!interface_info.initial_value.empty()) {
-          double value = hardware_interface::stod(interface_info.initial_value);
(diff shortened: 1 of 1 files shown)


gazebo_ros2_control/src/gazebo_system.cpp
```diff
@@ -233,14 +233,24 @@ void GazeboSystem::registerJoints(
...
+            initial_value = hardware_interface::stod(interface_info.initial_value);
+            RCLCPP_INFO(this->nh_->get_logger(), "\t\t\t found initial value: %f", initial_value);
+          } catch (std::invalid_argument &) {
+            RCLCPP_ERROR_STREAM(
+              this->nh_->get_logger(),
+              "Failed converting initial_value string to real number for the joint "
+                << joint_name
+                << " and state interface " << interface_info.name
+                << ". Actual value of parameter: " << interface_info.initial_value
+                << ". Initial value will be set to 0.0");
```
bmagyar, line 250: why should it be 0? if someone doesn't pay enough attention to the terminal, they may miss this message and continue operating with a broken system. If we know that the setup is broken, we should re-throw that exception after the log.

Wiktor-99: Okay, I don't have strong opinion on this. I will re-throw exception.

Wiktor-99:
@bmagyar As pipeline is fixed I assume that failed action cloud be trigger again.

ahcorde:
https://github.com/Mergifyio backport humble iron

PR #297, opened 2024-04-07
Rewrite mimic joints

### LINKED ISSUE #173: "_mimic" joint interface name suffix causes Moveit2 "Joint not found" Error
Is the mimic joint suffix that gets added to interface names ([here](https://github.com/ros-controls/gazebo_ros2_control/blob/6718ae9e60a43befc816968a2a1aa05395fc691b/gazebo_ros2_control/src/gazebo_system.cpp#L199)) necessary? If so, what is its intent? I can't find anything in the `ros2_control` source code that is expecting a mimic suffix, but maybe I'm not looking in the right place. 

I suppose it makes the output of services like `list_hardware_interfaces` a bit more verbose/explicit, but I'd argue that the onus should be on the URDF designer to decide whether the joint is named as a "mimic" in the name or not. As an aside, perhaps on the command interfaces output, `[mimic]` might be more appropriate than `[unclaimed]` and this is where that added string could go. 

I ask because I was trying to control a robot that had mimic joints with Moveit, simulated in gazebo, the Moveit RobotModel implementation complains loudly that `Joint 'joint_name_mimic' not found in model 'model_name'`, seemingly because the `GazeboSystem` `hardware_interface` assigns a new name to the joint that doesn't exist in the URDF/SRDF/yaml config files. It took me longer than I'd like to admit that the source of the error was `gazebo_ros2_control`'s change and not a mistake that I had made in one of the description files. 

Commenting out the ["_mimic" suffix]((https://github.com/ros-controls/gazebo_ros2_control/blob/6718ae9e60a43befc816968a2a1aa05395fc691b/gazebo_ros2_control/src/gazebo_system.cpp#L199)) doesn't appear to break anything, and allows Moveit to function as expected. 

Thanks in advance for any feedback/insights!

christophfroehlich:
The component_parser of the resource_manager now already parses the mimic joint information from the URDF. The old way for defining it in the `ros2_control` tag is deprecated, see https://github.com/ros-controls/ros2_control/pull/1256.

If this one is approved, I'll make the same changes for gz_ros2_control as well.

- This PR uses the info from the `hardware_info` structure instead of parsing it again.
- From my understanding, mimic means "mimic position", but the [URDF](http://wiki.ros.org/urdf/XML/joint) spec is not very clear in this point. So I implemented it that independently of the interface type of the mimicked joint, the mimic joint's position is set. See also https://github.com/ros-controls/ros2_control/pull/1497 for the discussion.
- replaces #111
- fixes #173 (remove suffix `_mimic`):
```bash
$ ros2 control list_hardware_interfaces
command interfaces
        right_finger_joint/position [available] [claimed]
state interfaces
        left_finger_joint/effort
        left_finger_joint/position
        left_finger_joint/velocity
        right_finger_joint/effort
        right_finger_joint/position
        right_finger_joint/velocity
```

image


doc/index.rst
```diff
@@ -101,35 +101,28 @@ include
...
-    <param name="mimic">right_finger_joint</param>
-    <param name="multiplier">1</param>
-    <command_interface name="position"/>
-    <state_interface name="position"/>
-    <state_interface name="velocity"/>
-    <state_interface name="effort"/>
-  </joint>
-
+The mimic joint must not have command interfaces configured in the ``<ros2_control>`` tag, but state interfaces can be configured.
+Independent of the interface type of the mimicked joint in the ``<ros2_control>`` tag, the mimic joint will use the position interface of the gazebo classic physic engine to follow the position of the mimicked joint.
```
saikishor, line 125:
```suggestion
**Note:** Independent of the interface type of the mimicked joint in the ``<ros2_control>`` tag, the mimic joint will use the position interface of the gazebo classic physic engine to follow the position of the mimicked joint.
```

gazebo_ros2_control/src/gazebo_system.cpp
```diff
@@ -340,6 +314,15 @@ void GazeboSystem::registerJoints(
     // check if joint is actuated (has command interfaces) or passive
     this->dataPtr->is_joint_actuated_[j] = (joint_info.command_interfaces.size() > 0);
   }
+
+  // set initial position for mimic joints
+  for (const auto & mimic_joint : hardware_info.mimic_joints) {
+    this->dataPtr->sim_joints_[mimic_joint.joint_index]->SetPosition(
+      0,
+      mimic_joint.offset + mimic_joint.multiplier *
+      this->dataPtr->joint_position_[mimic_joint.mimicked_joint_index],
```
saikishor, line 323:
```suggestion
      mimic_joint.offset + (mimic_joint.multiplier *
      this->dataPtr->joint_position_[mimic_joint.mimicked_joint_index]),
```

gazebo_ros2_control/src/gazebo_system.cpp
```diff
@@ -551,10 +534,11 @@ GazeboSystem::perform_command_mode_switch(
...
 
-  // mimic joint has the same control mode as mimicked joint
-  for (const auto & mimic_joint : this->dataPtr->mimic_joints_) {
-    this->dataPtr->joint_control_methods_[mimic_joint.joint_index] =
-      this->dataPtr->joint_control_methods_[mimic_joint.mimicked_joint_index];
+  // mimic joint has always position control mode
+  for (const auto & mimic_joint : this->info_.mimic_joints) {
+    this->dataPtr->joint_control_methods_[mimic_joint.joint_index] &=
+      static_cast<ControlMethod_>(VELOCITY & EFFORT);
+    this->dataPtr->joint_control_methods_[mimic_joint.joint_index] |= POSITION;
```
saikishor, line 541: @christophfroehlich can you explain me what you are doing here?

You are making sure that we only have the control method in position but not in velocity and effort?

saikishor (commented):
The PR overall looks good to me. Thanks for the changes. I left few comments for clarification

doc/index.rst
christophfroehlich: I formatted it as sphinx-note, see screenshot above.

saikishor: I like it. Thanks

gazebo_ros2_control/src/gazebo_system.cpp
christophfroehlich: Exactly, this is the code used elsewhere. I didn't manage to simply set it to `POSITION` first, because this is of type SafeEnum
`typedef SafeEnum<enum ControlMethod_> ControlMethod;`
but a `static_cast` seems to work now, see the last commit.

saikishor: Yes. Now it is very clear. Thanks

saikishor (commented):
LGTM. Thank you for the changes :)

saikishor (approved):

gazebo_ros2_control_demos/launch/gripper_mimic_joint_example_position.launch.py
```diff
@@ -0,0 +1,88 @@
+# Copyright 2022 Stogl Robotics Consulting UG (haftungsbeschränkt)
```
ahcorde, line 1:
```suggestion
# Copyright 2024 Stogl Robotics Consulting UG (haftungsbeschränkt)
```

doc/index.rst
```diff
@@ -250,19 +246,26 @@ When the Gazebo world is launched you can run some of the following commands to
   ros2 run gazebo_ros2_control_demos example_tricycle_drive
 
 
-The following example shows parallel gripper with mimic joint:
+The following example shows a parallel gripper with a mimic joint:
+
+
+.. code-block:: shell
```
ahcorde, line 252:
```suggestion
The following example shows a parallel gripper with a mimic joint:

.. code-block:: shell
```

ahcorde (changes requested):
Tested locally, LGTM, just two minor comments

ahcorde (approved):

PR #314, opened 2024-05-10
Deprecate gazebo_ros2_control on rolling

christophfroehlich:
Rolling CI fails because rolling is not released to jammy, but osrf/ros:rolling-desktop-full still uses jammy.

But even if it will be updated to noble it will fail because there is no gazebo classic release any more.


christophfroehlich:
@mergifyio backport humble iron

ahcorde (approved):

PR #329, opened 2024-05-18
[CI] Add workflow to test on humble

christophfroehlich:
I discussed with @saikishor that we want to keep the gazebo_ros2_control alive because the rolling version of gz_ros2_control can't be easily installed on jammy, see https://github.com/ros-controls/ros2_control_ci/pull/82

To test the rolling version here I re-add a workflow (deleted with #314), but test it on the humble distro by building the whole rolling ros2_control stack as we do [here](https://github.com/ros-controls/ros2_control_ci?tab=readme-ov-file#compatibility-versions)


saikishor (approved):
I agree with @christophfroehlich. It is better for now to have a tag in gazebo_ros2_control rolling that works on older distros. We would like to maintain the rolling version compilable and working for other users who would like to compile and release in their CIs for Humble and Iron. 

@ahcorde what do you think about this?

ahcorde (approved):

PR #357, opened 2024-07-08
Propagate the node clock and logging interface to ResourceManager

saikishor:
Needs: https://github.com/ros-controls/ros2_control/pull/1585

This changes helps to use the clock and logging interface of the node directly.


ahcorde (approved):

saikishor:
@ahcorde can you also tag the rolling branch of this repo as well. Anyway, we won't be making any releases so it shouldn't affect anyone.

